PR #4, opened 2020-05-29
Pre-amble and Interfaces

droyad:
Slack discussion: https://octopusdeploy.slack.com/archives/CTZT49JFJ/p1590559880133100

### DIFF
--- a/StyleGuide/Interfaces.md
+++ b/StyleGuide/Interfaces.md
@@ -1,0 +1,19 @@
+# Interfaces
+
+Interfaces are a level of indirection between the consumer and implementation. This indirection is often useful, but may sometimes be unnecessary. 
+
+## Dependency Injection
+
+Concrete implementations can be dependency injected. Therefor it is not necessary to create an interface for a type for the sole purpose of injection.
+
+## Location
+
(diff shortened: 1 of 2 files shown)


StyleGuide/Interfaces.md
```diff
@@ -0,0 +1,19 @@
+# Interfaces
+
+Interfaces are a level of indirection between the consumer and implementation. This indirection is often useful, but may sometimes be unnecessary. 
+
+## Dependency Injection
```
pawelpabich, line 5: This rule seems to be scoped to dependency injection only. What if we extended the scope? E.g.  

Try to follow YAGNI and avoid adding interfaces when they provide no immediate value.
* Concrete implementations can be dependency injected. Therefore it is not necessary to create an interface for a type for the sole purpose of injection.
* Other examples where we tend to use interfaces but we shouldn't

pawelpabich:
👍

StyleGuide/Interfaces.md
droyad: I had a bit of trouble thinking about what those other cases are

PR #7, opened 2020-06-01
Out/Ref parameters

droyad:


StyleGuide/OutAndRef.md
```diff
@@ -0,0 +1,18 @@
+
+## Out
+Avoid using `out` parameters except when there is a well established pattern in the .NET ecosystem, e.g. `Dictionary.TryGet`, `TryParse`. Even then the alternatives might be better.
+
+## Ref
+Avoid. At the time of writing, only 6 methods took `ref` parameters, so chances are your code doesn't need it. The .NET Framework only uses them in specific cases (e.g. `Interlocked`).
+
+## Alternatives
+
+The `Maybe<>` and `Result<>` types exist to indicate whether a result is returned and in the case of `Result<>` an error message if it failed.
```
slewis74, line 10: Would using nullable reference types be a better long term direction than `Maybe<>`?

droyad: Yes. Since nullable reference types only do something if the caller has it enabled, we should have a good portion of the code enabled before changing the recommendation.

matt-richardson: I'd like for us to completely move away from `Maybe<>` - the api on it is so confusing, it hinders so much more than it helps.

johnsimons: Same for `Map` and `Tee`

StyleGuide/OutAndRef.md
```diff
@@ -0,0 +1,18 @@
+
+## Out
+Avoid using `out` parameters except when there is a well established pattern in the .NET ecosystem, e.g. `Dictionary.TryGet`, `TryParse`. Even then the alternatives might be better.
```
johnsimons, line 3: What are the alternatives?

StyleGuide/OutAndRef.md
uglybugger: It's worth noting that `out` and `ref` parameters don't play nicely with async code. If we ever plan to go down that path, I suspect we'll have much more to deal with, but it's still worth keeping in mind.

uglybugger: Re `Maybe`: I agree it's confusing but I don't want to throw the baby out with the bath water. The idea of a discriminated union is very attractive, even if the `Maybe` implementation ended up morphing over time to something a little confusing.

PR #8, opened 2020-07-07
Architecture Decision Record: Config As Code API

andrewabest:
### Overview
This PR introduces an Arhcitecture Decision Record for the Config As Code Public API.

### Contributing
Read through the ADR document and associated documents linked, and if you wish to contribute to the decision, comment on this PR or in #topic-architecture, referencing this PR and linking to your thread in a comment here.

### DIFF
--- a/Decisions/20200707 - Config As Code API.md
+++ b/Decisions/20200707 - Config As Code API.md
@@ -1,0 +1,24 @@
+ - **Subject**: Public API Surface for Configuration As Code Feature
+ - **Decision date**: [when was the decision made]
+ - **Decision contributors**: Andrew Best, Paul Gradie, Adam McCoy, Matt Richardson, Andrew Moore, Rob Wagner
+ - **Decision maker**: TBA
+ - **Decision owner**: TBA
+
+Executive Summary: [Created once a decision is made]
+
+Options & Considerations: https://docs.google.com/document/d/1d9XBGd9akeHWn0B-sTPFV6KvfsR17abpBrHtoBRQadE
+
+Decisions: [what was the decision made]
+
+- Resource Identification:
+- Resource Location:
+- Resource Modification:
+- Retrieving VCS metadata:
+- Creating Releases:
+
+Revisions: 
+**[7/7/2020]**
+- Discussed the config-as-code spike carried out by Tribe Red with Matt Richardson and Andrew Moore. 
+- Arrived at the conclusion we want to discuss Resource Identification further
+- Agreed to an approach on Resource Location that would expose Hypermedia links off of the Branch resource
+- Confirmed approaches to modification, VCS metadata and Release creation implemented in the spike were appropriate (just needed productionising)
--- a/Decisions/20200707 - Config As Code API.md
+++ b/Decisions/20200707 - Config As Code API.md
@@ -1,8 +1,8 @@
- - **Subject**: Public API Surface for Configuration As Code Feature
- - **Decision date**: [when was the decision made]
- - **Decision contributors**: Andrew Best, Paul Gradie, Adam McCoy, Matt Richardson, Andrew Moore, Rob Wagner
- - **Decision maker**: TBA
- - **Decision owner**: TBA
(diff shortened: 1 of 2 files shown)


Decisions/20200707 - Config As Code API.md
```diff
@@ -0,0 +1,30 @@
...
+
+Decisions: [what was the decision made]
+
+- Resource Identification:
+- Resource Location:
+- Resource Modification:
+- Retrieving VCS metadata:
+- Creating Releases:
+
+Revisions:
```
droyad, line 19: I'm not sure what value this style of revision log is. It seems geared towards recording that some process was followed and feels very corporatey.

I'd rather see the options that were considered and why one was chosen over the others.  I'm not sure we need to record the timeline on how it played out.

andrewabest: It provides context as to what discussions were had and how the decision / options evolved over time. I don't see it as much different to a set of comments in a PR or a google doc, just centralised and summarised. One benefit I get from it here is whilst this PR stays open, it reminds me of the conversations I've had to date and smaller decision points hit for this particular document. 

Recording who we talked to isn't about corporate style 'accountability' - it is just more context. If we know who gave input later on, we can go back and talk to them about it if we need help in the future.

If we take an approach of detailling considered options and explaining the decision made, we might not be able to create an ADR until we have all of the information at-hand. That might be OK - as things evolve, we could keep it in a google doc, and not move to an ADR until we have finalised a decision. The downside to this is we don't capture any of the evolution. The upside is it is flexible, and allows our template to be more lightweight.

andrewabest: @michaelnoonan @slewis74 any chance either of you could provide a third view on the above? Looking to merge the base ADR template PR (https://github.com/OctopusDeploy/Architecture/pull/9) by the end of this week and would like to gain some concensus on the value of the revision log within it.

michaelnoonan: I advocate seeing a worked example with the revision log so we can comment on something concrete rather than imagined. @andrewabest it sounds like you have something useful in mind, and I'd like to see if it has value on reflection.

> the options that were considered and why one was chosen over the others
I would be sad if we lost clarity on this. I like the idea that someone can read these a year or so after the fact and gain most of the context they would have by being part of the original conversation.

andrewabest: That information is captured in the proposed template under `Options & Considerations` and `Decision`. 

@michaelnoonan there are a couple of revisions on this document just below to consider.

I think in the spirit of trying the simplest thing first, I'll remove the revisions and encourage this history is captured in the `Options & Considerations` section.

andrewabest: @droyad along the same lines - do we care to identify a decision maker / owner in these documents then, or just a list of people who had input to the decision?

pawelpabich: I decided to go with a little bit different approach in my ADR https://github.com/OctopusDeploy/Architecture/pull/10. 

> I decided to replace Revisions section with Data Sources section. I think that the vast majority of ADR consumers want to know what the current decision is.

>There will be a much smaller group of people who want to gain a deeper understanding of why the decision was made the way it was. Options & Considerations and Data Sources sections should be able to provide that detailed information.


Thoughts?

andrewabest: @pawelpabich I like that, I'll update the core template to reflect that approach.

droyad: I think who made the decision, owns it and who was involved is not important. We don't do this for bits of code. If someone really cares they can read through the comments and commits.

andrewabest: The information may get captured in the Data Sources @pawelpabich added to the base template. It seems to be pretty cheap to record who was involved in discussing / giving input into a given decision, vs having to go spelunking for that information / having it lost over time when you may need it. Happy to omit the maker / owner pieces - as you have mentioned, these are easily inferred by the author of the PR.

PR #10, opened 2020-07-16
Design Principles For Octopus Server Clients

pawelpabich:
I decided to replace `Revisions` section with `Data Sources` section.  I think that the vast majority of ADR consumers want to know what the current decision is. 

There will be a much smaller group of people who want to gain a deeper understanding of why the decision was made the way it was. `Options & Considerations` and `Data Sources` sections should be able to provide that detailed information.

### DIFF
--- a/Decisions/20200716-ClientDesignPrinciples.md
+++ b/Decisions/20200716-ClientDesignPrinciples.md
@@ -1,0 +1,53 @@
+ - **Subject**: Design Principles For Octopus Server Clients
+ - **Decision date**: TBD
+ - **Decision contributors**: Pawel Pabich, Shannon Lewis, Rob Wagner, Mike Noonan
+ - **Decision maker**: TBD
+ - **Decision owner**: School Of Architecture
+
+## Executive Summary
+
+ Clients should be as light as possible and the heavy lifting should be done by the server. 
+
+## Decision
+
+Keep the logic in the server. 
+
+If references to existing resources need to be passed to a command as one of its inputs then there should be a way to pass them either by `Id` or `Name`. Referring to resources by name might lead long term to a less stable solution but it helps our customers to get a working solution quickly.  
+
+Any errors should be reported back with enough details so the customer knows which inputs need to be modified. 
+
+## Options & Considerations: 
+
+### Keep the logic in the server
+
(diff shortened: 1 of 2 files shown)


Decisions/20200716-ClientDesignPrinciples.md
```diff
@@ -0,0 +1,53 @@
...
+ - **Decision owner**: School Of Architecture
+
+## Executive Summary
+
+ Clients should be as light as possible and the heavy lifting should be done by the server. 
+
+## Decision
+
+Keep the logic in the server. 
+
```
andrewabest, line 14: I'm wondering if we can format the below like this:

```
The only logic clients should be responsible for is:
- marshalling and sanitising user provided parameters
- allowing users to specify resource parameters by name or Id
  - If references to existing resources need to be passed to a command as one of its inputs then there should be a way to pass them either by `Id` or `Name`. Referring to resources by name might lead long term to a less stable solution but it helps our customers to get a working solution quickly.  
- reporting errors
  - Any errors should be reported back with enough details so the customer knows which inputs need to be modified. 
```

Decisions/20200716-ClientDesignPrinciples.md
```diff
@@ -0,0 +1,53 @@
...
+ - **Decision contributors**: Pawel Pabich, Shannon Lewis, Rob Wagner, Mike Noonan
+ - **Decision maker**: TBD
+ - **Decision owner**: School Of Architecture
+
+## Executive Summary
+
+ Clients should be as light as possible and the heavy lifting should be done by the server. 
+
+## Decision
+
```
andrewabest, line 12: Could we make this a bit more explicit? Something like "Keep all domain logic in the server".

Decisions/20200716-ClientDesignPrinciples.md
pawelpabich: 👍

Decisions/20200716-ClientDesignPrinciples.md
pawelpabich: 👍

andrewabest:
:shipit:

PR #11, opened 2020-07-21
API Versioning

droyad:


OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,109 @@
...
+The route itself will contain an additional version identifier after the resource name, e.g. `/api/Spaces-1/deploymentprocesses/v2` or `/api/Spaces-1/deploymentprocesses-v2`. Other approaches (e.g. parameters or header) do not fit with the hypermedia link approach.
+
+The client will pick the link knows about and wants to use. If the link is not available the client can fallback to a previous version, or present an informative error message. For clients not using the hypermedia, they will get a `404`.
+
+This option is as if we added a brand new route, allowing a total change of schema and behaviour. The `v1` API will retain it's behaviour, including limiting the records returned on a `list` to those previously understood (e.g. exclude `AwsFeed`). 
+
+The `v1` and `v2` schemas will be handled by different responders and different resource types. This will make it easy to remove `v1` when the time comes. Removal all also only affect those clients actively using that route.
+
+#### Other versioning schemes
+Passing the version via a query parameter (`&version=2.0.0`) or header (`X-Octopus-Api-Version: 2.0.0`) does not fit with the hypermedia link approach.
```
matt-richardson, line 74: Would be good to expand on this... I'm curious as to why you don't think the header approach is viable, especially seeing we mandate an `X-Octopus-Api-Key` header already?

droyad: Given a hypermedia document that differentiates between API versions using link names, there is no way to instruct the client to add a specific header.

matt-richardson: But there's no way to instruct it that `X-Octopus-Api-Key` should be added. Nor that `Content-Type` should be set to `application/json`?

droyad: How would the client know that for the `DeploymentProcesses` api it needs to pass `X-Octopus-Api-Version: 2.0.0` and for the `Events` api it needs to pass `X-Octopus-Api-Version: 3.0.0`? The client would need to be hard coded for this and, unless some other mechanism is added, doesn't know whether the server supports these new versions in advance.

matt-richardson: Wouldn't the client need that knowledge anyway? The code that calls the `DeploymentProcess` api would need to change anyway between v1 and v2 (otherwise, there'd be no point in changing the version number). "I'm expecting to send and receive v1 classes, so I'll use the v1 endpoint. I choose that by sending <xxx> header".

My preferred approach is to use the ContentType header to control this (though its not a hill I'm interested in dying on)... On request you send 
```
ContentType: X-Octopus-DeploymentProcess-V1
Accept: X-Octopus-DeploymentProcessResponse-V1
```
which determines the behaviour.

droyad: The client would gain that knowledge by the presence of a specific route in the root document. I haven't thought about the client side of this too much, and am open to this idea. I'll update the document.

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+One of the reasons we didn't adopt versioning was that there was no intention to remove old versions. This proposal implies we will remove them at some stage.
+
+### Implementation
+
+In the most common case we handle the requests and responses to be compatible with both the old (`v1`) and new (`v2`) schemas.
+
+We use some marker in the request to determine whether it is `v1` and map to `v2` if required. e.g. If we receive a request with `NuGetFeedId` we will map that value to `FeedId`.
+
+On response we write out the response in a way that is compatible with both schemas. e.g. the response will contain both `NuGetFeedId` and `FeedId`. For `list` we will return all records.
```
flin-8, line 41: Having the response contain both properties will be confusing, won't it?

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+This client would still need another way to indicate the expected version on a `get`.
+
+#### Conclusion
+This option would likely just be a complication on some sort of API versioning.
+
+### Whole of API
+
+The whole API is versioned in tandem, including the root document. To a client, the API at a specific version never changes. 
+
+This is easy to reason about and understand from a client perspective. If we made large but rare changes to the API, this would work well.
```
flin-8, line 93: If we went with root route (`/api/v2`), could we only bump the version for breaking changes (major bump)? Additions would be at the same version

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+The downside is that every route will  have a version bump regardless whether it changed, making it difficult to track which APIs have changed. Also if we retired an API version that an older client was using, it would break that API even though the client may have been fully compatible with the latest API.
+
+#### Conclusion
+
+This does not suit our cadence or the desire for not breaking clients without reason. We tend to make small and rare breaking changes.
+
+### Per Endpoint
+
+The API will be versioned per endpoint. An endpoint being a group of routes (and method) that handle the same shapped data. Example `/api/Spaces-1/feeds/all` and `/api/Spaces-1/feeds/{id}` would be grouped, but `/api/Spaces-1/feeds/summary` would not.
```
pawelpabich, line 103: I think we are redefining here a well-defined term here.  I would omit `endpoint` and use  `group of routes `.  Our `route` = `endpoint`

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+By adoption API versioning we will have the freedom to radically change the schema and behaviour of our endpoints (within the bounds of the backend storage considerations). 
+
+It would be as if we added a brand new endpoint. The `v1` API will retain it's behaviour, including limiting the records returned on a `list` to those previously understood (e.g. exclude `AwsFeed`). 
+
+We will deprecate and remove previous versions of an API. The exact mechanism is the subject of a different proposal.
+
+### Community Library
+
+Community library steps would be treated the same as if they were posted via the API. We would need to update the library and sync process to indicate the `DeploymentProcess` API version targeted.
```
matt-richardson, line 69: This is a good one to call out. It's going to be a challenge though, as we need to support many many versions of octopus with the sync process.

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Header
+
+The API will be versioned per endpoint `ContentType` and `Accept` headers. e.g:
+```
+ContentType: application/vnd.octopus.deploymentprocess.v2
+Accept: application/vnd.octopus.deploymentprocess.v3; q=1.0, application/vnd.octopus.deploymentprocess.v2; q=0.9
+```
+
+The correct handler will be picked based on the header value. If the server does not support the specified version, it should return `415 Unsupported Media Type` or `406 Not Acceptable`.
```
matt-richardson, line 145:
```suggestion
The correct handler will be picked based on the header value (in descending order of `q` value). If the server does not support any of the specified versions, it should return `415 Unsupported Media Type` or `406 Not Acceptable`.
```

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+The client would tell the server all the versions it accepts.
+
+#### Conclusion
+
+This fits better with the spirit of a Resource based API. However it does not allow browsing and is more difficult to understand.
```
matt-richardson, line 151: The swagger would (should?) expose this quite well.

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+
+The route itself will contain an additional version identifier after the resource name, e.g. `/api/Spaces-1/deploymentprocesses/v2` or `/api/Spaces-1/deploymentprocesses-v2`.
+
+The client will pick the link knows about and wants to use. If the link is not available the client can fallback to a previous version, or present an informative error message. For clients not using the hypermedia, they will get a `404`.
+
+This method also allows users to continue browsing the API via the browser.
+
+#### Conclusion
+
+**Recommend** This approach fits with out existing approach of API discovery. Clients can easily discover whether the version of the API they depend on still exists.
```
matt-richardson, line 135: Opening in a browser gives them part of the information... It also helps them figure out what a `GET` returns, but it doesn't help with what a `POST` requires. Feels like we're going to need to redirect them to the swagger for the full information.

OctopusServer/ApiVersioning.md
Waldo000000, line 69: > We would need to update the library and sync process to indicate the `DeploymentProcess` API version targeted.

Does this mean:
1. Update templates themselves to indicate the API version targeted (in which case, I wonder how we'd help library template authors determine this), and
2. Change the sync process logic (which currently just inserts a `CommunityActionTemplate` resource) to instead somehow follow the same code paths as are hit by requests to `POST /deploymentprocess`?

OctopusServer/ApiVersioning.md
droyad: Yes it is. This is what we currently do in a lot of places.

OctopusServer/ApiVersioning.md
droyad: Not quite, see https://github.com/OctopusDeploy/Architecture/pull/11/files#diff-6b209479df8a5f20a8081c61543ecf3eR76

OctopusServer/ApiVersioning.md
droyad: Yeh, this is a hidden problem we haven't really ever thought about. What happens if someone submits the new format to the library? It'll break on old servers. I think the library concern is related but not dependent on this proposal.

OctopusServer/ApiVersioning.md
droyad: I was using the term `Client` in terms of a software library. e.g. `Octopus.Client` could look for a link it knows about and then drive behaviour from there.

OctopusServer/ApiVersioning.md
droyad: The more I think about this particular problem, the more I think we need to re-think how we manage the library altogether (and version steps). I'm not sure it's dependent on this proposal at all although it quacks the same.

OctopusServer/ApiVersioning.md
matt-richardson: 🤔 This feels like there's a reasonable overlap with the [Design Principles For Octopus Server Clients proposal](https://github.com/OctopusDeploy/Architecture/pull/10).

droyad:
> Open Q: is our "recommended" method of api discovery still the swagger documenation?

For clients (the code kind) swagger hasn't been the recommended way, it uses the Hypermedia to figure out what the correct routes are.

OctopusServer/ApiVersioning.md
droyad: 👍  I've re-worded this, let me know if it needs further changes.

matt-richardson:
> > Open Q: is our "recommended" method of api discovery still the swagger documenation?
> 
> For clients (the code kind) swagger hasn't been the recommended way, it uses the Hypermedia to figure out what the correct routes are.

Hmm. Maybe "discovery" is the wrong word here. I mean "how does it work" / "what fields are there" / "what api endpoints are available?"
Might be worth calling out the distinction between "human discovery" and "does the server i'm talking to support this api".

matt-richardson:
> > Open Q: is our "recommended" method of api discovery still the swagger documenation?
> 
> For clients (the code kind) swagger hasn't been the recommended way, it uses the Hypermedia to figure out what the correct routes are.

Hmm. Maybe "discovery" is the wrong word here. I mean "how does it work" / "what fields are there" / "what api endpoints are available?"
Might be worth calling out the distinction between "human discovery" and "does the server i'm talking to support this api".

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,181 @@
...
+- Whether to make a change in the first place
+- Versioning of config as code documents
+- Versioning of step types
+- Policy and mechanics of the removal of old versions, other than it will be done
+
+## Background
+
+Up until now we have not versioned the API since `Server 3.0`. We've chosen between hard breaking the API contract, keeping it fully compatible and somewhere in between.
+
+One of the reasons we didn't adopt versioning was that there was no intention to remove old versions. This proposal implies we will remove them at some stage.
```
Waldo000000, line 33: I'm not sure that the usefulness of versioning depends on whether there's an intention to remove old versions.

This is because being explicit about which version a client is expecting to interact with enables us to respond with the expected resource shapes (rather than some ungodly combination of all possible shapes), and enables us to know for certain what the client's intent is when posting (rather than trying to sniff based on the shape of what they're posting).

OctopusServer/ApiVersioning.md
uglybugger: Some of the best API documentation I've encountered has been from [Stripe](https://stripe.com/docs/api/versioning). I'm tempted to add a vote for the root-level option, but not to force a lock-step version bump of all the routes.

droyad: >  I'm tempted to add a vote for the root-level option, but not to force a lock-step version bump of all the routes.
Could you describe how that would work? Isn't a root level version by definition all routes?

OctopusServer/ApiVersioning.md
droyad: Adding versioning adds new code paths, more code to maintain, multiple views over the same data and extra APIs to test.

Going on the Feed API example, since there was no intention of removing support for older clients, these code paths would have to be maintained indefinitely. This put weight on the argument to keep the single endpoint and keep it working transparently as it was felt as a better long term solution.

Waldo000000: > Adding versioning adds new code paths, more code to maintain, multiple views over the same data and extra APIs to test.

That's true. I think it's also true of the alternative approach. i.e. Faithfully supporting both the older and newer versions of some changed behavior still requires adding new code paths, new code to maintain, multiple views over the same data and extra test cases -- regardless of whether those two versions are _implicitly_ supported by a single endpoint or _explicitly_ supported by separate endpoints.

Unless you think this is important to flesh out though, I think we can jellyfish this for now as the historical reasoning is probably not a critical part of this proposal.

OctopusServer/ApiVersioning.md
pawelpabich: I agree that community library needs some love.  An author needs to get quite a few things right to submit a working template.  

That being said, templates can specify Min and Max version of Octopus Server they work with https://github.com/OctopusDeploy/OctopusDeploy/blob/ed78a72cfc03613d8c5dfc769bed81fcc24c00bd/source/Octopus.Server/Orchestration/ServerTasks/SynchronizeCommunityActionTemplates/Json/CommunityActionTemplateResource.cs#L30-L33

OctopusServer/ApiVersioning.md
uglybugger: I was thinking more along the lines of something like a "stable" API which is frozen after a period of time, and an "unstable" one which is permitted to vary without notice, and is also permitted to have incomplete coverage of the API surface. It would need to be completed by the time we mark it as stable, but until then we can just version-bump only the routes we need to materially change.

Potential scenario:

In our current situation:
`/api/thing/42` would be an alias for `/api/v1/thing/42` (i.e. all non-versioned API routes would forever be aliased to `v1`.
`/api/v1/thing/42` would be the "stable" version.
`/api/v2/thing/42` would be the "unstable" version. There might not be a `v2` version of all our routes at this point.

Time passes...

`/api/thing/42` is still an alias for `/api/v1/thing/42`
`/api/v1/thing/42` is still the "stable" version
`/api/v2/thing/42` has had a bunch of additions to it and we're considering making it the "stable" version, so we add all the missing routes for the unchanged parts of the API surface.

Time passes...

`/api/thing/42` is still an alias for `/api/v1/thing/42`
`/api/v1/thing/42` is "stable" but deprecated.
`/api/v2/thing/42` is "stable" and our recommended version.

Time passes...

`/api/thing/42` is still an alias for `/api/v1/thing/42`
`/api/v1/thing/42` is "stable" but deprecated.
`/api/v2/thing/42` is "stable" and our recommended version.
`/api/v3/thing/dc1bd5c5-d842-fda9-eb53-ef092bd6acb5/` is new and "unstable" and does not cover our entire API surface yet.

Time passes...

`/api/thing/42` is removed and now returns a 404.
`/api/v1/thing/42` is removed and now returns a 404.
`/api/v2/thing/42` is stable but deprecated.
`/api/v3/thing/dc1bd5c5-d842-fda9-eb53-ef092bd6acb5/` is stable and our recommended version..
`/api/v4/thing/<whatevz>` is unstable and does not cover our entire API surface yet.

... and we keep going up until we run out of 32-bit integers ;)

Does that make sense?

droyad: Makes sense

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### Conclusion
+
+With some clever pipeline logic we could make this transparent to the developer. The appeal is in the flexibility. It's not clear how well it would work with per-api versioning.
+
+## Options - version specifiers
+
+### No version
+
+If the client doesn't specify API version then we should fallback to the oldest supported version by the Server. This approach will keep existing clients as long operational as possible. There is a decent chance that the client won't be affected by the change. E.g. There is a new property in the property bag.
```
droyad, line 185: I think this is dangerous. We are changing the agreed upon contract with the client. Say we add a property in `v2`. The client, only understanding the `v1` schema, round trips the resource. The new property is now blanked out.

On the reverse, if we remove a property on `v2`, the client will send us the value and we will drop it silently.

Supporting `PATCH` operations or only making changes when a property is supplied (instead of assuming `null`) might get around this.

uglybugger: I think the underlying assumption of defaulting to the lowest supported version is a bit fraught. My thinking is that we're probably better off assuming v1 for all unversioned API calls, and eventually just deprecating and then 404ing for calls that don't specify a version.

pawelpabich: > Supporting PATCH operations or only making changes when a property is supplied (instead of assuming null) might get around this.

I agree.  I would expect a validation error when the new property is not supplied.  Assumptions around the lack of data in the request has bitten us quite a few times in the past. Sticking to "property present but set to null" as a way rule would be a good start.

pawelpabich: > I think the underlying assumption of defaulting to the lowest supported version is a bit fraught. 

I agree. The idea is to keep the existing gazillion lines of custom PS code working as long as possible.  Defaulting to `v1` is possibly an acceptable compromise between rejecting all calls on day one because they don't specify the version explicitly and accepting all calls unless they cause an error.

droyad: 👍 Intent with this was that anything without a version path would be `v1` and continue to work until removed.

pawelpabich: Also, we could make `v1` the default behaviour and provide customers with an option to switch to `lowest supported version`.  This will affect how the server treats requests without a version specified. 

This is similar to how we allow customers to disable `default space` so all clients are required to include `SpaceId` in all space-specific routes.

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,188 @@
...
+On response we write out the response in a way that is compatible with both schemas. e.g. the response will contain both `NuGetFeedId` and `FeedId`. For `list` we will return all records.
+
+Clients implemented similar logic to determine if it received a `v1` or `v2` response or whether posting a `v2` formatted is possible.
+
+### Applied to Scenarios
+
+Applying the scenarios to the current approach:
+
+- **Strict client** - The client deserializes the resource into a well known object structure, ignoring any fields it does not understand (e.g. C# client)
+- **Loose client** - The client either manipulates the JSON directly or deserializes into a dynamic structure, including all fields in the resource (e.g. Javascript, PowerShell, JObject)
```
andyinaus, line 50: Does this mean all the client API now take in sth like `JObject` as its parameter for a `Loose Client`? This may be a breaking change for C# client users.

OctopusServer/ApiVersioning.md
andyinaus: Would it make more sense to return `410 Gone` instead for those retired endpoints?

uglybugger: You're probably right. Good point :)

OctopusServer/ApiVersioning.md
droyad: No, we won't change how `Octopus.Client` works, however customers may be using `JObject`

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Implementation
+
+In the most common case we handle the requests and responses to be compatible with both the old (`v1`) and new (`v2`) schemas.
+
+We use some marker in the request to determine whether it is `v1` and map to `v2` if required. e.g. If we receive a request with `NuGetFeedId` we will map that value to `FeedId`.
+
+On response we write out the response in a way that is compatible with both schemas. e.g. the response will contain both `NuGetFeedId` and `FeedId`. For `list` we will return all records.
+
+Clients implemented similar logic to determine if it received a `v1` or `v2` response or whether posting a `v2` formatted is possible.
```
andrewabest, line 43: I'm not sure I understand why clients need to determine the response type - they should be able to assume they have recieved a response that aligns with the request version.

OctopusServer/ApiVersioning.md
andrewabest: If we map `v1`=>`v2` on the way in, is there a reason we would not map back on the way out? Thinking of the scenarios:

Rename: we map back to the v1 response
New Field: gets dropped
New Type: doesn't appear in list

My assumption is that if someone has specifically requested a `v1` endpoint, they will expect a `v1` response, and would also expect subsequent requests from the same client to be for `v1` endpoints (for the scenario where they retrieve a resource from one endpoint to provide as input to another).

This would save our read models (wire format / response models) becoming polluted across multiple versions.

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,188 @@
...
+| -------- | ------------- | ------------ |
+| Rename | Handled transparently | On `update` the client will send both properties, making it hard to discern intent if the conflict |
+| New Field | Field will be dropped on `update` | Handled transparently |
+| New Type | On `list` it may work or fail depending on the operation, possibly silently | On `list` it may work or fail depending on the operation, possibly silently
+
+## Outcome
+
+By adoption API versioning we will have the freedom to radically change the schema and behaviour of our endpoints (within the bounds of the backend storage considerations). 
+
+It would be as if we added a brand new endpoint. The `v1` API will retain it's behaviour, including limiting the records returned on a `list` to those previously understood (e.g. exclude `AwsFeed`).
```
andrewabest, line 63: I feel this doesn't align with what was stated above:

> On response we write out the response in a way that is compatible with both schemas

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Community Library
+
+Community library steps would be treated the same as if they were posted via the API. We would need to update the library and sync process to indicate the `DeploymentProcess` API version targeted.
+
+
+## SemVer
+
+We would use SemVer where applicable to version the API. A major increment indicates a breaking change in the contract (e.g. field removal, new type, rename), while a minor increment indicates an addition (e.g. new field). 
+
```
andrewabest, line 75: It is a concern that a minor change is actually a breaking change for some clients. They lose the 'semantic' part of it. 

I also wonder about how we will go about tracking / advertising what version of the API is shipped with what release of server. Is there any value to thinking about whether we could align the api version and the server version?

OctopusServer/ApiVersioning.md
```diff
@@ -0,0 +1,188 @@
...
+### Hypermedia Links
+
+The API will be versioned via hypermedia.
+
+In the case of per-endpoint versioning, the root document will contain additional links for each version. For example:
+```json
+"Links": {
+    "DeploymentProcesses":  "`/api/Spaces-1/deploymentprocesses`",
+    "DeploymentProcesses/v2":  "`/api/Spaces-1/deploymentprocesses-v2`",
```
andrewabest, line 123: This feels like it goes against the spirit of hypermedia - we increase the semantic barrier by making clients need to make decisions about a version within Links.

What if we drove the links from the requested version? The client has already indicated what version they want to work with upon issuing the request - via headers, querystring, doesn't supply and gets default, or whatever other mechanism we support. The Links collection can be populated with appropriate links as of that version. 

That means some links might be for `v2` if the endpoint has a new version available, and some will be for `v1` (and so on if more versions exist). 

This way we don't force clients to go through an additional layer of decision making to retrieve a link. They already indicated they wanted to work with `v2`, so the `DeploymentProcess` link should be the `v2` link.

OctopusServer/ApiVersioning.md
droyad: This section is describing what we currently do. The problems with this approach as you alluded to are that fields get dropped, as outlined in ` Applied to Scenarios` immediately below

OctopusServer/ApiVersioning.md
droyad: This is what currently happens. There is no request version, so the client has to deduce from the shape they got back.

OctopusServer/ApiVersioning.md
droyad: That's right because we want to move away from the problems it causes

OctopusServer/ApiVersioning.md
droyad: We don't want to force customers to upgrade all their integrations every time they upgrade the server. We aim for upgrades to be easy.

OctopusServer/ApiVersioning.md
andrewabest: Ah my apologies, I think I lost the context that this all sat under the `## Background` section. I wonder if updating `### Implementation` to `### Current Implementation` would help keep that context front of mind?

OctopusServer/ApiVersioning.md
droyad: That is assuming whole of API versioning isn't it? This section talks about the per-endpoint versioning option.

OctopusServer/ApiVersioning.md
andrewabest: As above, looks like I lost the context that we were discussing current state here, apologies 👍

OctopusServer/ApiVersioning.md
andrewabest: No, this would work with per-endpoint versioning.

It would look roughly like:

- Client X sends a request to /api to retrieve our root document, and specifies "V2" as the X-Octopus-Api-Version header.
- The Server constructs a root document with links to resources. The latest link is used for each resource. Some may have "V2" links, like our deployment process above. Those without a "V2" will have their latest link (V1, or 1.1, or whatever they have)
- This gets sent back to Client X. The client wants to retrieve deployment processes, so it accesses the "DeploymentProcesses" link, and issues a request to the V2 endpoint.
- The processes returned will also have "Self" links to the V2 endpoint, as the subsequent request will also have the same V2 Api-Version header.

So the server constructs appropriate hypermedia per-resource based on the known requested API version (or the default if none was requested)

OctopusServer/ApiVersioning.md
andrewabest: Totally agree.

I don't necessarily see this as requiring that though - Stripe for example use a date for their API version. But it isn't like you need to always update the date on your integrations whenever they add a new version.

I'm not convinced using a date or the server version is a good idea, more thinking out aloud here - if we versioned the API with the server version for example, we would need to record what changed in each version for the API, in a similar way to our current release notes. 

Integrations would indicate what version of server they were designed to work with. If changes arrived people wanted to access, they would update their server version, and the version of their integrations to match. Our infrastructure that mapped from one version to another could be aware of these version numbers and be driven from them.

Thoughts?

Waldo000000:
@droyad @uglybugger Has this been superceded by the [#project-api-philosophy proposal](https://docs.google.com/document/d/1uEueYqeRSEheh2qJUUSFpdHcepAgnD7x0uM24eqzEyE/edit#heading=h.xzo1xyjv96da)? Or is the document in this PR consistent with and complementary to it?

uglybugger:
> @droyad @uglybugger Has this been superceded by the [#project-api-philosophy proposal](https://docs.google.com/document/d/1uEueYqeRSEheh2qJUUSFpdHcepAgnD7x0uM24eqzEyE/edit#heading=h.xzo1xyjv96da)? Or is the document in this PR consistent with and complementary to it?

It's been rolled into the proposal from #project-api-philosophy. Mostly consistent, but a few ideas have been fleshed out a little better. Still not ready for prime-time, though, as we still need to accommodate OpenAPI-generated clients etc. which don't understand hypermedia.

PR #14, opened 2020-07-22
Technical Debt

droyad:
**Readable Version**: https://github.com/OctopusDeploy/Architecture/blob/technicalDebt/Decisions/TechnicalDebt.md
**Timeline**: No hurry, aim to close this by the end of this cycle (Cycle 4)

## Description
This was spurred on by recent debate about whether we should overload a term or make a breaking change. This document is based on the thought that we should treat technical debt as a line of credit. It's ok to incur debt as long as you pay some of it down, but not necessarily the same debt.

- Comment on individual lines
- Comment on the PR. Consider using long form replies detailing your position.
- Add 👍 comments or positions you support, it will be tallied when gauging degree of consensus.

### DIFF
--- a/Decisions/TechnicalDebt.md
+++ b/Decisions/TechnicalDebt.md
@@ -107,7 +107,7 @@ This allows us to meet a customer requirement without planning too far into the 
 
 After consultation, an individual or team introduces a new architecture, pattern or design. 
 
-The limit it to one section of the application. This may be because they are not sure the pattern is useful, or because there is not enough time.
+They limit it to one section of the application. This may be because they are not sure the pattern is useful, or because there is not enough time.
 
 The code is left with two ways of doings things.
 
--- a/Decisions/TechnicalDebt.md
+++ b/Decisions/TechnicalDebt.md
@@ -78,7 +78,7 @@ It is impossible to predict the future, but we can take an informed guess at it.
 
 We may never pay off low interest debt.
 
-## Which debt should be pay down?
+## Which debt should we pay down?
 
 We should be paying down the highest interest debt first. 
 
--- a/Decisions/TechnicalDebt.md
+++ b/Decisions/TechnicalDebt.md
@@ -1,11 +1,3 @@
-# Overview
-
-- **Subject**: Whether and how much technical debt we take on and how we pay it back
-- **Decision date**: [when was the decision made]
-- **Decision contributors**: [list who was involved in the decision]
-- **Decision maker**: Engineering Managers
-- **Decision owner**: [who executes on the final decision]
-
 # Executive Summary
 
 Octopus will maintain a revolving line of technical credit. We borrow against future earnings into order to move quicker. We will purposefully incur technical debt as well as pay it down indefinitely. 
@@ -19,7 +11,7 @@ As the company and it's install base has grown the technical debt is starting to
 
 # Approach
 
-We accept that we will have an ongoing level of technical debt. The level will shift as the priorities of the product change. The engineering managers are responsible for monitoring this.
+We accept that we will have an ongoing level of technical debt. The level will shift as the priorities of the product change. The L5s in engineering with the help of L4 are responsible for monitoring this.
 
 Engineers are empowered to incur debt in order to ship features. They should do so deliberately and thoughtfully.
 
--- a/Decisions/TechnicalDebt.md
+++ b/Decisions/TechnicalDebt.md
@@ -115,7 +115,7 @@ Whether to remove it, gradually spread it, or undertake a pitch to complete it i
 
 ## Outgrown
 
-As the product grows and the userbase changes, previous assets may suddenly become liabilities. The incurred debt could not have reasonably forseen, or if it was it was very distant and nebulous. 
+As the product grows and the userbase changes, previous assets may suddenly become liabilities. The incurred debt could not have been reasonably foreseen or, if it was, it was very distant and nebulous. 
 
 This is unavoidable.
 
--- a/Decisions/TechnicalDebt.md
+++ b/Decisions/TechnicalDebt.md
@@ -1,0 +1,129 @@
+# Overview
+
+- **Subject**: Whether and how much technical debt we take on and how we pay it back
+- **Decision date**: [when was the decision made]
+- **Decision contributors**: [list who was involved in the decision]
+- **Decision maker**: Engineering Managers
+- **Decision owner**: [who executes on the final decision]
+
+# Executive Summary
+
+Octopus will maintain a revolving line of technical credit. We will purposefully incur technical debt as well as pay it down. 
+
+# Background
+
+For the first 7 years of it's life, Octopus was a very small company. It managed to produce a great product and do so quickly. Part of this was deliberately taking on technical debt 
+and design patterns in order to move faster <sup>[citation needed]</sup>. 
+
+As the company and it's install base has grown the technical debt is starting to slow us down noticeably. It's no longer possible for engineers to learn the whole product and where the sharp edges are. This has cause development to slow down and a reduction in reliability.
+
+# Approach
+
+We accept that we will have an ongoing level of technical debt. The level will shift as the priorities of the product change. The engineering managers are responsible for monitoring this.
+
+Engineers are empowered to incur debt in order to ship features. They should do so deliberately and thoughtfully.
+
+When taking on debt we do not need a plan for how to pay down that specific debt. This will allow us to move faster and often these things are clearer in hindsight.
+
+We will have an indefinite program of paying down technical debt. Currently ~30% of pitch capacity each cycle.
+
+# FAQ
+
+## What is technical debt?
+
+`Technical Debt` is an overloaded term, but at Octopus it encompasses:
+- A lean approach to automated testing focusing on the green path in favour of manual testing
+- Hairy code that does the job but is hard to understand and modify
+- Code and designs that are bug prone
+- Monolithic spaghetti design
+- The UI, API and DB not agreeing on what a thing is called (ubiquitous language)
+- Architecture and designs that have we have outgrown
+- Partially completed refactors
+- Multiple patterns for doing the same thing
+
+`Technical Debt` is **not**:
+- Inconstancies in how features work
+- Missing features
+- Bad UX
+- Bugs (although it may cause bugs)
+
+We use `interest` to describe the consequence arising from the debt, mainly in lost time and customer good will.
+
+## How is technical debt introduced?
+
+Debt can be introduced a number of ways, not limited to:
+- Shortcuts to ship a pitch
+- Backwards compatibility or versioning requirements
+- The trial of a new code pattern
+- Lack of knowledge of existing patterns
+- The thousand cuts of minor additions and fixes
+
+## Is technical debt bad?
+
+No. As with financial debt, healthy businesses take on liabilities in order to make higher revenue with that the increase in revenue exceeds the repayment on the debt. Businesses also often operate on a revolving line of credit, they never intend to repay the principal.
+
+To apply this to code, a certain level of debt is desired. Paying it all down is counter productive as there is something more valuable that could have been done instead. If Octopus Server ever reaches it's end of life, the debt will evaporate.
+
+## How much technical debt should we have?
+
+> Perfect is the enemy of good - Voltaire
+
+The exact quantity depends on the observer. Not so much that it slows us down significantly and causes many reliability issues. However not so little that we ship no features.
+
+## Is all technical debt equal?
+
+[This article](https://digitalassetmanagementnews.org/opinion/the-technical-debt-metaphor-a-better-alternative/) reframes technical debt as `Call Options`.
+
+It is impossible to predict the future, but we can take an informed guess at it. Code that will likely never change, or change rarely (and is read rarely) would incur less interest than code that is on the main code path and subject to frequent change. For example team management vs the step editor. 
+
+We may never pay off low interest debt.
+
+
+# Examples
+
+## API Compatibility and Versioning
+
+As part of a pitch the team decides they need to maintain backwards compatibility. They choose to write code to keep it compatible. For example:
+- Re-use an field or term, changing it's meaning without changing it's name (breaking ubiquitous language)
+- Modify the existing endpoint to be backwards compatible
+- Add a new versioned endpoint
+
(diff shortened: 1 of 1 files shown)


Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,129 @@
...
+
+When taking on debt we do not need a plan for how to pay down that specific debt. This will allow us to move faster and often these things are clearer in hindsight.
+
+We will have an indefinite program of paying down technical debt. Currently ~30% of pitch capacity each cycle.
+
+# FAQ
+
+## What is technical debt?
+
+`Technical Debt` is an overloaded term, but at Octopus it encompasses:
```
droyad, line 34: Should we be using a different term? `Technical Liability`?

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+- A lean approach to automated testing focusing on the green path in favour of manual testing
+- Hairy code that does the job but is hard to understand and modify
+- Code and designs that are bug prone
+- Monolithic spaghetti design
+- The UI, API and DB not agreeing on what a thing is called (ubiquitous language)
+- Architecture and designs that have we have outgrown
```
Waldo000000, line 40:
```suggestion
- Architecture and designs that we have outgrown
```

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+- Partially completed refactors
+- Multiple patterns for doing the same thing
+
+`Technical Debt` is **not**:
+- Inconstancies in how features work
```
Waldo000000, line 45:
```suggestion
- Inconsistencies in how features work
```

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+Debt can be introduced a number of ways, not limited to:
+- Shortcuts to ship a pitch
+- Backwards compatibility or versioning requirements
+- The trial of a new code pattern
+- Lack of knowledge of existing patterns
+- The thousand cuts of minor additions and fixes
+
+## Is technical debt bad?
+
+No. As with financial debt, healthy businesses take on liabilities in order to make higher revenue with that the increase in revenue exceeds the repayment on the debt. Businesses also often operate on a revolving line of credit, they never intend to repay the principal.
```
Waldo000000, line 63:
```suggestion
No. As with financial debt, businesses take on liabilities in order to make higher revenue. This can be healthy, as long as the increase in revenue exceeds the extra repayments on those liabilities. Businesses also often operate on a revolving line of credit, that is, they never intend to fully repay the principal.
```

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+
+To apply this to code, a certain level of debt is desired. Paying it all down is counter productive as there is something more valuable that could have been done instead. If Octopus Server ever reaches it's end of life, the debt will evaporate.
```
Waldo000000, line 65:
```suggestion
Applying this same logic to engineering implies that a certain level of debt is desired. Paying it all down is counterproductive when there is something more valuable that could have been done instead. If Octopus Server ever reaches its end of life, the debt will evaporate.
```

michaelnoonan:
Overall, I like this. It resonates with my experiences.

I mostly liked the concrete examples - they are easy for me to agree: "Yes, I've seen this before."

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+# Background
+
+For the first 7 years of it's life, Octopus was a very small company. It managed to produce a great product and do so quickly. Part of this was deliberately taking on technical debt 
+and design patterns in order to move faster <sup>[citation needed]</sup>. 
+
+As the company and it's install base has grown the technical debt is starting to slow us down noticeably. It's no longer possible for engineers to learn the whole product and where the sharp edges are. This has cause development to slow down and a reduction in reliability.
+
+# Approach
+
+We accept that we will have an ongoing level of technical debt. The level will shift as the priorities of the product change. The engineering managers are responsible for monitoring this.
```
michaelnoonan, line 22: Managers only? Is it more an expectation of Level 4-5 across the team? Looking at Octopus People, we only talk about technical debt in Levels 2-3, but it seems like a really important subject for Level 4-5.

For example: I want a Level 4-5 to provide actionable insights about inhibitors/debts as they go about their work. Actionable meaning we have a concrete plan to pay down that particular debt.

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Is all technical debt equal?
+
+[This article](https://digitalassetmanagementnews.org/opinion/the-technical-debt-metaphor-a-better-alternative/) reframes technical debt as `Call Options`.
+
+It is impossible to predict the future, but we can take an informed guess at it. Code that will likely never change, or change rarely (and is read rarely) would incur less interest than code that is on the main code path and subject to frequent change. For example team management vs the step editor. 
+
+We may never pay off low interest debt.
+
+## Which debt should be pay down?
```
Waldo000000, line 81:
```suggestion
## Which debt should we pay down?
```

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+
+We should be paying down the highest interest debt first. 
+
+It is often not obvious what is the best way to pay down newly acquired debt. It is also not clear how much interest it will change. It's only after some time that the cost and the way forward becomes clearer.
```
Waldo000000, line 85:
```suggestion
It is often not obvious what is the best way to pay down newly acquired debt. It is also not clear how much interest it will charge. It's only after some time that the cost and the way forward becomes clearer.
```

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+
+No. As with financial debt, businesses take on liabilities in order to make higher revenue. This can be healthy, as long as the increase in revenue exceeds the extra repayments on those liabilities. Businesses also often operate on a revolving line of credit, that is, they never intend to fully repay the principal.
+
+Applying this same logic to engineering implies that a certain level of debt is desired. Paying it all down is counterproductive when there is something more valuable that could have been done instead. If Octopus Server ever reaches its end of life, the debt will evaporate.
+
+## How much technical debt should we have?
+
+> Perfect is the enemy of good - Voltaire
+
+The exact quantity depends on the observer. Not so much that it slows us down significantly and causes many reliability issues. However not so little that we ship no features.
```
pawelpabich, line 71: This is what was mentioned right at the beginning:

> We will have an indefinite program of paying down technical debt. Currently ~30% of pitch capacity each cycle.

I think we should come up with a number so we can keep ourselves honest. Continuing with financial debt as an analogy, households should not spend more than 30% of their income on their mortgage. Should we aim at 30%?  

How does QRF fit into that?   E.g. Lack of a data migration caused X long and thus expensive investigations in QRF. 

E.g. If we dedicate 5 people working in features + 4 people in QRF to deal with the technical debt then thats  42% (9 / 21) of engineering.

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+
+This debt is taken on with the understanding that we will consider whether and how to remove it at a future date.
+
+This allows us to meet a customer requirement without planning too far into the future. 
+
+## Architectural Trials
+
+After consultation, an individual or team introduces a new architecture, pattern or design. 
+
+The limit it to one section of the application. This may be because they are not sure the pattern is useful, or because there is not enough time.
```
Waldo000000, line 110:
```suggestion
They limit it to one section of the application. This may be because they are not sure the pattern is useful, or because there is not enough time.
```

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+As part of a pitch the team decides they need to maintain backwards compatibility. They choose to write code to keep it compatible. For example:
+- Re-use an field or term, changing it's meaning without changing it's name (breaking ubiquitous language)
+- Modify the existing endpoint to be backwards compatible
+- Add a new versioned endpoint
+
+In doing so they introduce additional code paths or inconsistencies that require maintenance and may confuse future developers.
+
+They add metrics to measure the usage of the old code paths.
+
+This debt is taken on with the understanding that we will consider whether and how to remove it at a future date.
```
Waldo000000, line 102: Or, perhaps, with the understanding that we _immediately_ flag/communicate what the blockers were, and start conversations about what underlying engineering efforts might be worth investing in, to avoid those blockers recurring for future Pitches?

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+
+The limit it to one section of the application. This may be because they are not sure the pattern is useful, or because there is not enough time.
+
+The code is left with two ways of doings things.
+
+Whether to remove it, gradually spread it, or undertake a pitch to complete it is left for the future.
+
+## Outgrown
+
+As the product grows and the userbase changes, previous assets may suddenly become liabilities. The incurred debt could not have reasonably forseen, or if it was it was very distant and nebulous.
```
Waldo000000, line 118:
```suggestion
As the product grows and the userbase changes, previous assets may suddenly become liabilities. The incurred debt could not have been reasonably foreseen or, if it was, it was very distant and nebulous. 
```

Decisions/TechnicalDebt.md
droyad: 30% to pay down the technical debt principal. Long term this may be a little high, but right now at a stage where the interest has increased due to our growth.

Interest is payed by extra support required, effort to fix things arising due to debt and hinderences in getting things done.

In your analogy your mixing the payment of principal with payment of principal and interest. Households aim to pay their principal to 0 because they want to retire. Businesses usually carry debt forever.

I don't think we should link the debt to the number of people in engineering. The debt has allowed us to grow revenue faster.

I think the decision should be based on whether money is better spent paying down debt or writing new features. Say each cycle we lose 0.5 people cycles due to paying interest on debt X. If it costs us 2 people cycles to pay that debt down, we'd recoup our effort in 4 cycles. However if we only lose 0.1 people cycles and it costs 5 people cycles, it'll take 50 cycles to get the benefit.

It should be reasonably straight forward to quantify the amount of debt. The interest might be a bit more tricky, there are so many aspects, slower development, bugs, customer perception, etc. Gut feel is that interest payments are way to high.

Decisions/TechnicalDebt.md
```diff
@@ -0,0 +1,136 @@
...
+- Partially completed refactors
+- Multiple patterns for doing the same thing
+
+`Technical Debt` is **not**:
+- Inconsistencies in how features work
+- Missing features
+- Bad UX
+- Bugs (although it may cause bugs)
+
+We use `interest` to describe the consequence arising from the debt, mainly in lost time and customer good will.
```
droyad, line 50: I think we should highlight the distinction better. You could have a huge mountain of debt, but it's not hindering, therefor low interest (e.g. current API design). Conversely you could have a small debt that has huge interest (e.g. the one function that causes a lot of bugs).

I think it's `interest` we should track as a metric, and think about the right level of that and which debt to paydown to best reduce that interest.

Decisions/TechnicalDebt.md
mcasperson: I think technical debt is synonymous with phrases like “code smell”. The mental energy I devote to a conversation around the technical debt of code I have written is unreasonably high, if only to convince myself that my technical debt doesn’t make me unworthy of the job. If I experience that with many years of experience, non-native English speakers, new hires or those starting their engineering careers must also feel the weight of a term like technical debt.

This document does a wonderful job of explaining the nuance of what Octopus means by technical debt. But I’d argue the term itself (ironically) carries to much social debt.

I think something like “technical bet” describes what we are trying to achieve here.

JeremyCade: `Debt` and `Liability` are not synonymous. 
In terms of business and accounting, they have well-defined meanings. 

- `Liability` is any obligation that has arisen from an event (e.g. a Transaction, Tax Liability). 
- `Debt` is a sub-class of Liability that has a well-defined term of payment (e.g. Borrowed Money). 
- All Debts are Liabilities.
- Not all Liabilities are Debts.

andrewabest:
@droyad what is the intent of this document? Why would I come and look for it? How would it help me in making a decision or taking some sort of action?

droyad:
@andrewabest The intent is to help teams make good decisions and do so independently while they are building a feature. It gives permission for the team to increase the technical debt in order to deliver or perform an experiment. It also give assurances to those teams that we will at the same time continually pay down the technical debt so it doesn't get out of hand.

It is a recognition that technical debt has been and continues to be a tool we hold strategically, that we are aware of it and that we are deliberate about it.

Some scenarios are outlined in the [examples](https://github.com/OctopusDeploy/Architecture/pull/14/files#diff-2755678f1f320a1d00bdba56a08188bdR89).

Some examples of the :
- The team spends the whole 6 weeks refactoring in order to avoid adding debt
- Improvements are not made because they are to large to do in one go, and to do so incrementally would increase debt in the short term
- We do not take risks as debt will increase if they don't pay off

The incident that sparked this document was the decision about how to handle the change of the `JSON Variable Replacement` feature into a more general `File Variable Replacement` feature. i.e should they:
- Keep the existing name and just make an additive change
- Rename the feature it to keep the ubiquitous language but make a breaking change
- Have both features running side by side
- Have some sort of compatibility thing

My initial reaction was "don't introduce more tech debt" as we're trying to pay it down. This would have involved significantly more work. However we came to the conclusion that not all debt is the same and we can both increase debt while decreasing other debt.

andrewabest:
@droyad thanks for the clarification. 

This document feels very akin to the quality document in confluence: https://octopushq.atlassian.net/wiki/spaces/IN/pages/1031209000/Improving+Quality+in+2020#Cheatsheet - setting expectations and giving people permission to invest in quality as they go about their work. 

One of the nice things this document does is clearly outlines the problem https://octopushq.atlassian.net/wiki/spaces/IN/pages/1031209000/Improving+Quality+in+2020 to frame the intent of the document. I wonder if something similar could be incorporated into this document?

I quite like the concrete example you have provided here - do you think we could work this real example into the examples section?

andrewabest:
Please ensure this PR conforms with the new guidance on structure in https://github.com/OctopusDeploy/Architecture/blob/master/README.md

PR #15, opened 2020-07-23
Document the Result object pattern we are using in the libraries

slewis74:


matt-richardson:
Can we put this proposal out for review before merging?

PR #16, opened 2020-07-24
Removing Functionality

droyad:


OctopusServer/FunctionalityRemoval.md
```diff
@@ -0,0 +1,135 @@
...
+
+We do this by deprecating features and giving customers some time to transition over before eventually removing it. We develop features and procedures that assist customers in performing the transition.
+
+# Overview
+
+This document focuses on when we remove functionality from Octopus Server and the upgrade implications for customers
+
+It does not cover
+- How to decide whether to deprecate and when to remove
+- Which functionality will help the user make the transition
```
droyad: This will be detailed in a pitch, currently [here](https://docs.google.com/document/d/1Jo8cxYoqiNK1HqgO4jayUTHcXs4etWbunQPA_faSkPw/edit#heading=h.g165efrlab8g)

OctopusServer/FunctionalityRemoval.md
```diff
@@ -0,0 +1,135 @@
...
+- Deprecated features hanging around for years
+- Old conversion and compatibility code hanging around
+- An ever-increasing combination of Octopus and integration versions we need to support.
+- Paralysis in the introduction of new features that break existing functionality
+
+[Draft 1](https://docs.google.com/document/d/1JFgWg4T6yTjU1epz85bHHMaU73tQ43bggsq8tRFIaU8/edit#heading=h.b13zj4am1nqv) of this proposal with comments
+
+# Breaking Changes Policy
+
+## Breaking Releases
```
droyad: I toyed with calling them `checkpoint releases`, but I removed it for now.

droyad:
I don't think the Adhoc option gives us a big advantage is is confusing for everyone.

The twice a year option feels good to me as then we can quickly remove things and not pay to maintain it. However on balance I think we should go with the `Once per year, shortened transition period`.

This would also reduce the number of steps a customer who is say 3 years behind. They would only need to do 3 installs.

We could communicate this quite clearly to customers by saying:
> If you are on a version earlier than `.4.0`, please upgrade to the latest version for that year, and then to the latest version in the next year and so on.

Not as clear as this, but a good compromise:
> You must upgrade major by major release.

OctopusServer/RemovingFunctionality.md
```diff
@@ -0,0 +1,153 @@
...
+This means we only need to keep the deprecated functionality for a small amount of time. It also avoids a "big bang" release with lots of changes.
+
+Example, given a desired transition period of 6 months (3 releases):
+| Deprecated In | Removed In | Transition Period |
+| ------------- | ---------- | ----------------- |
+| 2020.4.0 | 2021.1.0 | 6 months |
+| 2020.4.1 | 2021.2.0 | 8 months |
+| 2021.1.0 | 2021.4.0 | 6 months |
+
+### Option - Once per year
```
NickJosevski, line 82: I like the simplicity of this option 👍

droyad:
`Once per year, shortened transition period` seems to be the winner so far

OctopusServer/RemovingFunctionality.md
```diff
@@ -0,0 +1,153 @@
...
+Examples:
+| Deprecated In | Removed In | Transition Period |
+| ------------- | ---------- | ----------------- |
+| 2020.4.0 | 2021.1.0 | 6 months |
+| 2020.4.1 | 2022.1.0 | 18 months |
+| 2021.1.0 | 2022.1.0 | 12 months |
+| 2021.1.1 | 2023.1.0 | 12 months |
+
+#### Option - Twice per year
+
```
andrewabest, line 118: I like this option the most. It provides more optionality around when to deprecate and remove rather than only having a fixed once-per-year drop.

More importantly it keeps the transition period reasonable, for a couple of reasons:
- If you give customers 24 months to do something, they will do the thing in 23 months time (if at all)
- Tracking the work of deprecation and removal over a 24 month period, when we work in 6 week cycles, seems impractical - people will lose context of the changes, and actually shipping the removal will become more difficult.

OctopusServer/RemovingFunctionality.md
andrew-moore-octo: I like this option too 👍 This is easy for customers (and us) to remember and reason about.

OctopusServer/RemovingFunctionality.md
```diff
@@ -0,0 +1,153 @@
...
+- An ever-increasing combination of Octopus and integration versions we need to support.
+- Paralysis in the introduction of new features that break existing functionality
+
+[Draft 1](https://docs.google.com/document/d/1JFgWg4T6yTjU1epz85bHHMaU73tQ43bggsq8tRFIaU8/edit#heading=h.b13zj4am1nqv) of this proposal with comments
+
+# Approach
+
+## Breaking Releases
+
+We designate certain releases to be breaking releases well in advance. All planned breaking changes are made in this release.
```
andrewabest, line 41:
```suggestion
We designate certain releases to be breaking releases well in advance, according to our agreed schedule. All planned breaking changes are made in this release.
```

johnsimons:
My first gut feeling is that the more time we give to people the longer they take to execute on it and quite possible forget to do it.

johnsimons:
Also not clear on how the whole process works in conjunction with our 6 weeks cycles. Example, we decide to introduce feature X to replace feature Y, so we mark feature Y as obsolete, but the cycle is finished, who/how do we clean-up in x months?

johnsimons:
The final question is regarding Cloud instances.
So the on-prem offer is up to the users to decide when they ready to migrate and update, but the Cloud offer is a bit different, we "force" the users to upgrade. So what happens if users do not do it within the time allocated?

OctopusServer/RemovingFunctionality.md
```diff
@@ -0,0 +1,152 @@
...
+
+We designate certain releases to be breaking releases well in advance, according to our agreed schedule. All planned breaking changes are made in this release.
+
+## Transition Period
+
+We want to make the transition away from deprecated functionality smooth for customers. When we deprecate, we set the transition period by nominating the release it will be removed in.
+
+The transition period should be minimum of 6 months, which also aligns with our support schedule.
+
+When deprecating we add alerting and reporting to notify the user they are using deprecated functionality. We also add functionality and documentation to help the user to transition.
```
johnsimons, line 49: Would this mean the user would have to update Octopus to receive this information or are we going to push this info to them via eg OctopusJS?

droyad: They would have to update Octopus. This is because the detection code would be integrated into the C# code paths.

droyad:
> who/how do we clean-up in x months

We would schedule in engineering pitches to do this

droyad:
>  So what happens if users do not do it within the time allocated?

Things break for them. But we would have emailed them and notified them multiple times.

OctopusServer/RemovingFunctionality.md
droyad: My thought is that there would be a PR created at the same time to do the removal. It'll decay, but at least it'll give a head start.

I agree on the time. The problem is that customer lag quite significantly in their upgrades.

johnsimons:
I wonder if all removal is equal, eg.
If I am removing some feature without giving a replacement then yes a longer period may be required.
But on the other hand, if we are in essence renaming an API, so there is a replacement for it, maybe a shorter period is OK too.

PR #20, opened 2020-08-25
Domain Modelling in Octopus Server

andrewabest:
**Readable Version:** https://github.com/OctopusDeploy/Architecture/blob/adr-domain-model/Decisions/20200825-DomainModel.md

**Timeline:** Aiming to have concensus on the approaches presented by end of sharpening (Friday 28 Aug).

## Description
This PR builds on work completed by @johnsimons, @andyinaus and I that gives us the ability to use ASP.NET Core controllers, and move away from our legacy Nancy infrastructure. Nancy is no longer actively maintained, so this is a desirable direction for Octopus Server to take.

Instead of looking to replicate what we have in "the new world" of ASP.NET, we want to take the time to re-think some of our core architectural approaches before we start creating or migrating new endpoints in earnest.

What this PR presents is a new approach to modelling our core domain, which emphasises making Octopus Server easier to reason about. It does so primarily by being stricter with our resource type definitions and creation, and by enforcing invariants within our core model, as opposed to invariant enforcement being the shared responsibility of many disparate rules and services.

The options and opinions presented in this PR are represented in [this Octopus Server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834).

How to engage with this PR:
- Comment on individual lines
- Comment on the PR. Consider using long form replies detailing your position. Look for **FEEDBACK WANTED** annotations if you would like to understand the type of feedback we are directly looking for (although all feedback is welcome)
- Add 👍 comments or positions you support, it will be tallied when gauging degree of consensus.

I will leave this PR up for comment until **Friday 11 September.**

### DIFF
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -25,6 +25,9 @@ New components introduced in the worked example provided in this document sit on
 **Resource Mapping**
 This document sees us move from utilising the ResourceMapper to directly facilitating Resource to Domain Model mapping, but it does not directly address moving away from the mapper, or moving away from it when mapping from Domain Model => Resource when translating back to wire format on "the way out". This will be the focus of [another decision record]() TBA.
 
+**Domain Events**
+Domain event infrastructure helps us create rich, decoupled domain models that can evolve and scale in healthy ways. This PR does not introduce any new domain event infrastructure or considerations, although it is a topic we want to tackle soon after agreeing on the core domain approach and abstractions.
+
 ## Considerations
 
 [This PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834) works through an end-to-end example of refactoring the `Tenants` `PUT` and `POST` endpoints to use new commponents that align with the above principals.
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -28,7 +28,7 @@ This document sees us move from utilising the ResourceMapper to directly facilit
 **Domain Events**
 Domain event infrastructure helps us create rich, decoupled domain models that can evolve and scale in healthy ways. This PR does not introduce any new domain event infrastructure or considerations, although it is a topic we want to tackle soon after agreeing on the core domain approach and abstractions.
 
-## Considerations
+## Overview
 
 [This PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834) works through an end-to-end example of refactoring the `Tenants` `PUT` and `POST` endpoints to use new commponents that align with the above principals.
 
@@ -71,7 +71,7 @@ To achieve this, we use:
 
 - Encapsulation within our core models, to ensure they remain internally consistent.
 
-## Options
+## Considerations
 
 ### Resource Validation
 
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -11,8 +11,6 @@ TBA
 
 ## Principals
 
----
-
 - We want our core model to be simpler to reason about
 - We want our core model to be safer to interact with
 - We want less bugs that were caused by misunderstanding of what state the core model might be in
@@ -21,8 +19,6 @@ TBA
 
 ## Exclusions
 
----
-
 **ASP.NET Core**
 New components introduced in the worked example provided in this document sit on new ASP.NET Core infrastructure. Specifics around the best abstractions for controllers, or other ASP.NET Core pipeline specific considerations, are outside of the scope of this document.
 
@@ -30,8 +26,6 @@ New components introduced in the worked example provided in this document sit on
 This document sees us move from utilising the ResourceMapper to directly facilitating Resource to Domain Model mapping, but it does not directly address moving away from the mapper, or moving away from it when mapping from Domain Model => Resource when translating back to wire format on "the way out". This will be the focus of [another decision record]() TBA.
 
 ## Considerations
-
----
 
 [This PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834) works through an end-to-end example of refactoring the `Tenants` `PUT` and `POST` endpoints to use new commponents that align with the above principals.
 
@@ -75,8 +69,6 @@ To achieve this, we use:
 - Encapsulation within our core models, to ensure they remain internally consistent.
 
 ## Options
-
----
 
 ### Resource Validation
 
@@ -142,13 +134,9 @@ Feedback will be provided by two mechanisms:
 
 ## Decision
 
----
-
 TBA
 
 ## Data Sources
-
----
 
 Source listed below might provide additional context but keep in mind that they can disappear at any time.
 
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -109,9 +109,11 @@ We have broken the resource model for `Tenant` into two components:
 
 Example: [TenantResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L14-L36) vs [TenantReadResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L38-L76).
 
-There are some basic convention tests established in server within `ResourceConventionsFixture` to ensure `WriteResource`s remain true subsets of `ReadResource`s.
+There are some basic convention tests established in server within `ResourceConventionsFixture` to ensure `WriteResource`s remain true subsets of `ReadResource`s, to ensure the common simple integration pattern of GET an octpus resource, modify it, PUT it back\_ will still be valid.
 
-This aligns with the direction our web client is taking, which currently has the concept of a `TNewResource`. This uses typescript's `Omit` to remove properties from the full resource definition that are not needed when creating resources. We could align our client resources by providing changing `TNewResource` to `TWriteResource`. Modify methods could accept `TWriteResource & TLinks`, which read resources retrieved from the server should match anyway thanks to the ✨of duck-typing, making round-tripping resource changes simple.
+Separating our resource model between read and write resources aligns with the direction our web client is taking, which currently has the concept of a `TNewResource`. `TNewResource` implementations use typescript's `Omit` to remove properties from the full resource definition that are not needed when creating resources.
+
+We could align our web client resources by providing changing `TNewResource` to `TWriteResource`. Modify methods could accept `TWriteResource & TLinks`, which read resources retrieved from the server should match thanks to the ✨of duck-typing, ensuring round-tripping resource changes in our web client remains straightforward.
 
 **FEEDBACK WANTED** is this a sensible first step, or should we jump in with both feet and introduce seperate models for each endpoint to completely decouple their concerns?
 
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -104,12 +104,14 @@ Given our ideal approach to resource modelling uses nullable reference types and
 
 We have broken the resource model for `Tenant` into two components:
 
-- `TenantResource` is used to create or update tenants, and does not contain Id or SpaceId properties (as these are available as query parameters), or hypermedia links
+- `TenantWriteResource` is used to create or update tenants, and does not contain Id or SpaceId properties (as these are available as query parameters), or hypermedia links
 - `TenantReadResource` is used to provide tenant information to clients - it contains all tenant information, along with identifiers and hypermedia links
 
 Example: [TenantResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L14-L36) vs [TenantReadResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L38-L76).
 
-This aligns with the direction our web client is taking, which has the idea of a `TNewResource` which uses typescript's `Omit` to remove Id and Links properties from the full resource definition.
+There are some basic convention tests established in server within `ResourceConventionsFixture` to ensure `WriteResource`s remain true subsets of `ReadResource`s.
+
+This aligns with the direction our web client is taking, which currently has the concept of a `TNewResource`. This uses typescript's `Omit` to remove properties from the full resource definition that are not needed when creating resources. We could align our client resources by providing changing `TNewResource` to `TWriteResource`. Modify methods could accept `TWriteResource & TLinks`, which read resources retrieved from the server should match anyway thanks to the ✨of duck-typing, making round-tripping resource changes simple.
 
 **FEEDBACK WANTED** is this a sensible first step, or should we jump in with both feet and introduce seperate models for each endpoint to completely decouple their concerns?
 
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -107,7 +107,7 @@ We have broken the resource model for `Tenant` into two components:
 - `TenantResource` is used to create or update tenants, and does not contain Id or SpaceId properties (as these are available as query parameters), or hypermedia links
 - `TenantReadResource` is used to provide tenant information to clients - it contains all tenant information, along with identifiers and hypermedia links
 
-[Example](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#diff-51caff6d0c4dcb6968bdcf3f0ea615dd).
+Example: [TenantResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L14-L36) vs [TenantReadResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L38-L76).
 
 This aligns with the direction our web client is taking, which has the idea of a `TNewResource` which uses typescript's `Omit` to remove Id and Links properties from the full resource definition.
 
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -125,6 +125,8 @@ This means we are not militant in applying precondition checks to incoming infor
 
 The other type of precondition check we are typically required to do relates to enforcing relational consistency when linking documents to other documents, which is still a necessary requirement in the new world.
 
+To fulfill this type of precondition, we will use a simple `Func of IO` approach as demonstrated [here](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#r479883974), passing the domain model a function which it can invoke to retrieve the information it needs to inspect and satisfy it's own invariants.
+
 When interacting with the domain, if a domain invariant is unsatisifed [example](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#diff-29fae3c0e603cce39e381b742617dc85R120), we will throw a domain exception from the domain model, which can be captured and handled centrally in our `ErrorHandlingMiddleware`.
 
 **FEEDBACK WANTED** are we going to get an acceptable amount of correctness with this approach? What could bite us if we aren't applying precondition checks like `!= null` and `string.isNullOrEmpty` within our domain model?
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -125,9 +125,7 @@ The other type of precondition check we are typically required to do relates to 
 
 When interacting with the domain, if a domain invariant is unsatisifed [example](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#diff-29fae3c0e603cce39e381b742617dc85R120), we will throw a domain exception from the domain model, which can be captured and handled centrally in our `ErrorHandlingMiddleware`.
 
-We have also introduced a lightweight `IQuery<TRequest, TResult>` abstraction to make it simple to define small, composable queries that can be exposed on relevant `I{Resource}DocumentStore`s and leveraged to enforce domain invariants.
-
-**FEEDBACK WANTED** are we going to get an acceptable amount of correctness with this approach? What could bite us if we aren't applying precondition checks like `!= null` and `string.isNullOrEmpty` within our domain model? Does the query abstraction provide value?
+**FEEDBACK WANTED** are we going to get an acceptable amount of correctness with this approach? What could bite us if we aren't applying precondition checks like `!= null` and `string.isNullOrEmpty` within our domain model?
 
 ### Client Feedback
 
--- a/Decisions/20200825-DomainModel.md
+++ b/Decisions/20200825-DomainModel.md
@@ -9,6 +9,10 @@ TBA
 
 # Detail
 
+## Assumptions
+
+- all new endpoints created in Octopus Server will be ASP.NET Core controllers, and we will be actively migrating away from our legacy Nancy responder infrastructure. Nancy is no longer actively maintained, so this is a desirable direction for Octopus Server to take.
+
 ## Principals
 
 - We want our core model to be simpler to reason about
@@ -20,7 +24,7 @@ TBA
 ## Exclusions
 
 **ASP.NET Core**
-New components introduced in the worked example provided in this document sit on new ASP.NET Core infrastructure. Specifics around the best abstractions for controllers, or other ASP.NET Core pipeline specific considerations, are outside of the scope of this document.
+New components introduced in the worked example provided in this document sit on new ASP.NET Core infrastructure. Specifics around the best abstractions for controllers, the migration strategy from Nancy to ASP.NET Core, or other ASP.NET Core pipeline specific considerations, are outside of the scope of this document.
 
 **Resource Mapping**
 This document sees us move from utilising the ResourceMapper to directly facilitating Resource to Domain Model mapping, but it does not directly address moving away from the mapper, or moving away from it when mapping from Domain Model => Resource when translating back to wire format on "the way out". This will be the focus of [another decision record]() TBA.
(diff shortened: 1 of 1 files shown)


Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+Given our ideal approach to resource modelling uses nullable reference types and ASP.NET's awareness of them during model binding, we have taken a step toward _Asymmetric Resource Models_.
+
+We have broken the resource model for `Tenant` into two components:
+
+- `TenantResource` is used to create or update tenants, and does not contain Id or SpaceId properties (as these are available as query parameters), or hypermedia links
+- `TenantReadResource` is used to provide tenant information to clients - it contains all tenant information, along with identifiers and hypermedia links
+
+[Example](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#diff-51caff6d0c4dcb6968bdcf3f0ea615dd).
+
+This aligns with the direction our web client is taking, which has the idea of a `TNewResource` which uses typescript's `Omit` to remove Id and Links properties from the full resource definition.
```
TomPeters, line 108: I would like to see us align our naming pattern between the web client and backend. I don't have a strong opinion so happy to follow whatever pattern we end up adopting here.

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+
+**FEEDBACK WANTED** is this a sensible first step, or should we jump in with both feet and introduce seperate models for each endpoint to completely decouple their concerns?
```
TomPeters, line 110: >is this a sensible first step

:+1:

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+
+The other type of precondition check we are typically required to do relates to enforcing relational consistency when linking documents to other documents, which is still a necessary requirement in the new world.
+
+When interacting with the domain, if a domain invariant is unsatisifed [example](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#diff-29fae3c0e603cce39e381b742617dc85R120), we will throw a domain exception from the domain model, which can be captured and handled centrally in our `ErrorHandlingMiddleware`.
+
+We have also introduced a lightweight `IQuery<TRequest, TResult>` abstraction to make it simple to define small, composable queries that can be exposed on relevant `I{Resource}DocumentStore`s and leveraged to enforce domain invariants.
+
+**FEEDBACK WANTED** are we going to get an acceptable amount of correctness with this approach? What could bite us if we aren't applying precondition checks like `!= null` and `string.isNullOrEmpty` within our domain model? Does the query abstraction provide value?
```
TomPeters, line 124: >Does the query abstraction provide value?

While the pattern seems fine, I don't think the `IQuery` interface itself adds value (or at least I don't understand the value)

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Client Feedback
+
+When something goes wrong processing a request, we need to return feedback to clients so they might act upon it.
+
+Feedback will be provided by two mechanisms:
+
+- _Model Binding_ - when verifying the correctness of client payloads a response will be returned that contains all of the errors detected within the payload as a single response. This will allow clients to address all of these errors before proceeding. This is particularly useful in the Octoups web app.
+
+- _Domain Exceptions_ - when a domain invariant is unsatisfied, an exception will be thrown immediately, and processing of the given request will cease. This will mean these types of errors will be revealed "one per request". Typically we would not expect sets of this type of error. The client will have to correct the invariant, and then can attempt to process their request again.
```
TomPeters, line 134: One thing to consider: If these exceptions originate from our domain, then we only have access to the properties that exist within our domain. If we raise an error saying that property `Foo` is invalid, yet the matching property on our resources is actually called `Bar` (or any other number of ways this mapping isn't 1:1), then the user gets a bad experience.

This is not a big problem now because _most_ of the properties in our domain model match the resource properties. I think this is true because
1. it is simpler to understand when the properties match
2. our automatic property matching strategy used by `ResourceMapper` encourages us to follow this convention

If we do decide to change our resource mapping design (for example, to use hand crafted mapping functions), then point 2 may no longer be true. In such a world, an innocent rename refactor (that otherwise might not change behaviour) might have a difficult to predict effect on the UX when one of these invariants is violated.

Might be worth spending a few moments thinking about this problem at this junction if you haven't done so already.

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+
+**FEEDBACK WANTED** do we think this will provide an adequate client experience? Are there areas of the application where there could be a series of domain invariants that would be unsatisifed where this approach would not be sufficient?
```
TomPeters, line 136: It might also be worth spending a few moments thinking about how we might want to provide multiple domain errors per request in future, if we do decide to make that change. While I think one error per request is fine right now :+1: it would be nice if we could verify that the direction we are taking won't inhibit us from making a "multiple domain errors per request" direction change later on.

TomPeters:
General direction here looks good to me. Consider me a :+1: on each of the main points. 

I don't want to be too deeply involved in the finer details, or block any decision making, so I've just left a few key thoughts that were on my mind as I read this. Do with them as you will :)

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Options
+
+### Resource Validation
+
+A number of options for resource validation were considered, considering the following constraints:
+
+- We want to use nullable reference types and TinyTypes throughout Octopus Server
+- We need to work with ASP.NET model binding
+- We use JSON.NET for serialization
```
matt-richardson, line 82: Is this the new microsoft json library, or the newtonsoft one?

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+- Encapsulation within our core models, to ensure they remain internally consistent.
+
+## Options
+
+### Resource Validation
+
+A number of options for resource validation were considered, considering the following constraints:
+
+- We want to use nullable reference types and TinyTypes throughout Octopus Server
+- We need to work with ASP.NET model binding
```
matt-richardson, line 81: nit: is "need" the right word here? Or is it "we want to use ASP.NET model binding (as we want to move away from Nancy)"?

Would be good to work in this part of the PR description:
> [this] gives us the ability to use ASP.NET Core controllers, and move away from our legacy Nancy infrastructure. Nancy is no longer actively maintained, so this is a desirable direction for Octopus Server to take.

Decisions/20200825-DomainModel.md
matt-richardson: 🤔How does this cross over with backwards compat? If we explicitly have separate models, we can be more "liberal in what we accept, but strict in what we return" (Postel's law i think?).
A good example here of one to think about is the feed type - we [return different data](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Server/Web/Api/Feeds/FeedReMapper.cs#L9) depending on the caller. I think there are at least a handful of places where what we accept from the client is "interesting" because we support backwards compat there. Different models might play nicely with [api versioning](https://github.com/OctopusDeploy/Architecture/pull/11) proposal.
Not sure I have a strong opinion here - just raising for awareness.

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+
+To achieve this, we use:
+
+- Nullable reference types, to ensure we only store information that is appropriate for the model.
+
+- Tiny types, which help strongly align requirements for data between our resources and domain model.
+
+- Encapsulation within our core models, to ensure they remain internally consistent.
+
+## Options
```
matt-richardson, line 74: nit: I'm not sure this section is outlining options here. Consider dropping this heading.

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,146 @@
...
+This progressive enhancement does not gel with nullable reference types, which have a strict requirement to initialize non-null fields in constructors (or explicitly set them to null, which is yuck).
+
+Given our ideal approach to resource modelling uses nullable reference types and ASP.NET's awareness of them during model binding, we have taken a step toward _Asymmetric Resource Models_.
+
+We have broken the resource model for `Tenant` into two components:
+
+- `TenantResource` is used to create or update tenants, and does not contain Id or SpaceId properties (as these are available as query parameters), or hypermedia links
+- `TenantReadResource` is used to provide tenant information to clients - it contains all tenant information, along with identifiers and hypermedia links
+
+[Example](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834/files#diff-51caff6d0c4dcb6968bdcf3f0ea615dd).
```
matt-richardson, line 106: Nit: this example isn't very clear as to what it's telling me. Consider:
```suggestion
Example: [TenantResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L14-L36) vs [TenantReadResource](https://github.com/OctopusDeploy/OctopusDeploy/blob/0047fb8c59f47c9661c911329fbbdd624f7cc026/source/Octopus.Core/Resources/TenantResource.cs#L38-L76).
```

Looking at that file in isolation, I'm not sure I'm enthusiastic. It feels like we're going to easily get out of sync there. How can we make it easier to fall into the pit of success here?

Decisions/20200825-DomainModel.md
matt-richardson: My gut feel is that multiple domain errors at once will be nicer, but not mandatory

matt-richardson (commented):
👍 Broadly in agreement - nicely done.
I've added a few "nit" comments that may improve the overall readability - leave it to you to decide to 🙈 or update.
Also added a couple of bits of food for thought.

Decisions/20200825-DomainModel.md
johnsimons, line 82: Given we are trying to start from a green fields perspective it may make sense to so with the `System.Text.Json`. As it stands MS has acquired James so I am pretty sure Newtonsoft Json has its days counted and we don't want to have to worry about migrating off it in a years time.

Decisions/20200825-DomainModel.md
johnsimons, line 110: I thought there is a distinction between "create new" and "update" (We use the `CreateOnly` attributes), does this mean we would need 3 distinct resource types?

Decisions/20200825-DomainModel.md
andyinaus: If we wish to return multiple domain errors, we may also need to consider the following:
1. Each invalid request is now an expensive operation as most of the domain rules require accessing the data access layer to validate e.g. Does `SpaceId` refer to a real space? Is DeploymentTarget scoped correctly with the Account/Certificate scoping it is using? 
2. Domain errors could have dependencies on each other e.g. If users try to create a tenant in a space that doesn't exist, we wouldn't be able/may not want to validate whether the tenant tags and project environments provided exist or not.

Decisions/20200825-DomainModel.md
andrewabest: I can change it to 'want' - but rolling your own binding in aspnet core is pretty left-field, I've never seen it done before. I've seen custom model binders, but I've never seen anyone completely bypass model binding and roll their own 'read the request and hydrate the thing we are working with' around it.

Will adapt your second point, good call.

matt-richardson: Or is it "We need to use ASP.NET model binding as we want to use ASP.NET Core controllers"?

andrewabest: I've tried to weave this point a little more clearly. At the top I've added an assumption that we will be creating all new endpoints in ASP.NET Core and migrating from nancy. In this particular context (Request validation), that means we want to work with model binding as we will be working with controllers. Hopefully that clarifies it.

Decisions/20200825-DomainModel.md
andrewabest: The goal here isn't having our write and read models in sync, it is to have clear contracts for both, rather than a shared contract that is partially hydrated depending on who is holding it for what reason. We also have the constraint of strict nulls wanting to fully initialize any given model from the constructor.

That being said, if our `ReadResource` representations are really supposed to be 1:1 reflections of our stored domain model, perhaps one way to ease the burden would be to look at source generators for our `ReadResource` models + associated mappers?

Decisions/20200825-DomainModel.md
andrewabest: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#specify-constructor-to-use `System.Text.Json` doesn't support constructor initialization yet, so won't play as nicely with our usage of nullable reference types. I think we should hold off a migration until it does, as it allows us to define safer, more correct `Resource` models.

Decisions/20200825-DomainModel.md
andrewabest: I'm trying to find an example of where we do this @johnsimons (I can't find that attribute) - could you point me to one please?

Decisions/20200825-DomainModel.md
andrewabest: I agree that would be nice, but there is a difference between the two. On the web client, our `TNewResource` omits Id / Links, but the full resource is used for `PUT` / updates. On the server, we use a single resource for *writes*, and one for *reads*. I'm not sure if this can be easily aligned.

TomPeters: All we need to pass to the repository for updates is `TWriteResource & TLinks`. I.e. we pass the write resource through, plus we also need its links to get the URL (thanks to duck typing the full read resource is likely to match this).

Decisions/20200825-DomainModel.md
andrewabest: At the moment the change proposed is entirely backwards compatible - we no longer expect an Id or Links in our serialized model on the server, so if existing clients send them, we will just drop the values on the floor. They need to supply the Id in the route for `PUT`'s anyway.

I like to think that this behaviour is usually the extent to adapting Postel's law / robustness principle in good API design - ignore values you don't understand from clients, but be strict about the values you do know about.

If we find during the process of writing new endpoints that we want to split resource models in a way that isn't backwards compatible, I think that is where we should dip into the api versioning proposal and try align how we make those changes (including routes, hypermedia, etc) with that proposal.

johnsimons: Sorry @andrewabest it is actually the `Writeable` attribute I was referring to, eg https://github.com/OctopusDeploy/OctopusDeploy/blob/b6f984bce39dc9f2f818f68d3303bd10a113e423/source/Octopus.Core/Resources/ActionTemplateParameterResource.cs#L22

slewis74: See https://github.com/OctopusDeploy/OctopusDeploy/blob/cb9b364472968b0a56a7eafda031fa75a009d659/source/Octopus.Core/Resources/ReleaseResource.cs#L67
You can only set the ProjectId when creating a release, you can't change it on update.

Decisions/20200825-DomainModel.md
andrewabest: The two main things it gives us are:

- something to enforce consistently-shaped query components. They take an input, they give an output, they have an execute method. Consistent shapes make the codebase easier to grok and work with
- something that allows us to scoop all queries up and register them with our container, avoiding needing to register them one by one

The alternative would be building these queries as methods on the `IDocumentStore` in question, or just forgoing the interface.

I don't like the former - my subjective opinion is that we should be encouraging smaller components rather than ever-growing big ones.

The latter will lead to inconsistency propogating throughout the codebase, as it tends to rapidly when you have lots of engineers working on a thing and have no convention tests or abstractions in place.

What would your preference be for this type of querying?

Decisions/20200825-DomainModel.md
slewis74: I'm also curious how this will play out in the client. In a scripting world customers will be accustomed to calling GET, changing a field on the resource and PUT it back. If we separate the resources on that side so it's clear what is required and what cannot be updated it would make scripting much easier to reason about, but there is a lot of mapping between resource types to be done in that model.

Decisions/20200825-DomainModel.md
andrewabest: Good point! 

I'd hope the types of invariants that need to map 1:1 with a field name will be enforced at the client validation layer - model binding, and use the resource type naming. This is where at the server we say "you were supposed to give us a thing with an X on it, but we can't see X" - and hopefully the UI has a field named X and we are all happy.

Things coming out of the domain I would hope would refer to specific things (ideally by name), and be less about "This field is missing or the wrong shape". This might happen when we can't locate a package, or a thing, and so the message will be more about that than a particular field selection, so hopefully the UX holds up.

I could be wrong, this is just an assumption. 

There is also an incoming proposal to move toward rolling our own mappers as opposed to auto mapping that Andy will be putting up later this week 👍

Decisions/20200825-DomainModel.md
andrewabest: Let's look at the design at a high level:

- With the current design in Octoups Server, we recieve resources, we might validate them upon recieving them from a client, we then map them to our persistence model, and THEN we validate whether the result represents a valid state. To allow this to work, our domain needs to allow itself to be mutated into an invalid state. Nothing prevents us from then persisting this state too.

- With the proposed design, we recieve resources, we validate them upon recieving them from a client, and we then conduct a set of operations on our domain. If any of these operations fail, the domain will immediately throw an exception, ceasing processing, and ensuring no invalid state is committed. The domain will never let itself be put itself into an invalid state, giving us very strong guarantees about its invariants.

For any given complex operation, if we wanted to see out of the set of operations that comprise it which left invariants unsatisifed, we would need a domain service to take the responsibility of evaluating correctness of the operation's manifest / command model before interacting with the domain model (which would still enforce its own invariants too). 

This separation of responsibilities is perfectly OK - in fact most domain invariants will likely span a number of different aggregates, and so require a coordinating domain service to enforce correctness, rather than it entirely being enforced within an aggregate / model. 

We may choose that this enforcement live in the controllers, it could be pushed into command handlers if we decide the controllers are getting too fat, or we could create single-purpose domain services for each concern. There are plenty of options.

The Tl;Dr: is we aren't painting ourselves into a corner with this approach.

Decisions/20200825-DomainModel.md
andrewabest: :nice:

We might need to conduct a vote to see if we prefer explicit `TWriteResource` / `TReadResource` naming, or to have one be the canonical `TResource` and have the other be the `T{Write/Read}Resource`.

Decisions/20200825-DomainModel.md
andrewabest: Newtonsoft.

Decisions/20200825-DomainModel.md
andrewabest: If write resources remain true subsets of read resources, clients should still be able to do that - GET the resource, modify it, PUT it back.

As for fields that can be set on create but shouldn't change on update, there are a couple of options:

The first, and simplest, is we ignore the fields in update handlers. No more auto-mapping (subject of another proposal) will mean what fields we consume and set in our domain operations are entirely under our control.

The second, which becomes more complex, is that we seperate the models again - one for create, one for update, where appropriate. This allows us to be nice and explicit about the payload, and avoid confusing consumers (I sent that field, but it didn't change!). If we can ensure all write models remain true subsets of our read model for all resource-based interactions, happy days! We can enforce this with type based convention tests if we want. I may attach one to that PR actually.

There will also be domain operations modelled in our API that aren't just resource-based operations. It is OK that these don't directly map to resources / models - they are modelling specific operations, and we have a model in place to allow this modelling, [enveloped resources](https://github.com/OctopusDeploy/Architecture/blob/master/Decisions/20200803-EnvelopeResources.md)

Decisions/20200825-DomainModel.md
TomPeters: >Consistent shapes make the codebase easier to grok and work with

This is the key point I was missing, thanks! Personally I don't think that this level of consistency helps me too much more than a simple implicit convention would. I.e. if something diverges from the convention, it probably wouldn't confuse me, yet the interface itself did initially have me confused! I don't know if I'm an outlier or not :)

>something that allows us to scoop all queries up and register them with our container, avoiding needing to register them one by one

Personally, I think of the only entry point (and the API to the outside world) being the methods on the document stores - the queries themselves feel like an implementation detail that the document store manages. If the query is simple enough for it to be an inline implementation on the document store (no extra class) - that's fine! If it is complex enough to extract out into another class, that's fine too. In the latter case, I don't (currently) see too much benefit in needing to inject (and therefore register) these queries. It feels like the coupling between these queries and the document store is very high and they should never be used outside the document store, so I would be comfortable with the document store newing them up as needed.

I acknowledge that some of my assumptions could be wrong, or the level of complexity that we ultimately end up with could be large enough to warrant some DI to help us manage it.

Decisions/20200825-DomainModel.md
TomPeters: I'm thinking about backwards compatibility in our typed clients (our c# client). In this case, it would make sense to change the type of resource we expect for create/modify calls to the new thinner type, and depending on how we do this, it might break existing clients.

We could avoid making this change, and keep the c# client as is (since it would be backwards compatible with the API itself), yet this loses some of the other benefits of this proposal, such as helping our customers fall into the pit of success: If I'm a new customer automating the creation of a project for the first time, wouldn't it be amazing if the type system told me that the only things I need to provide are the name, LifecycleId, and ProjectGroupId, and it wouldn't even let me provide things like the SpaceId?

Similarly, while we could add overloads to maintain backwards compatibility (support both the thicker and thinner types for creating/updating resources), this has a tradeoff of making the API more confusing for new people - which overload are they supposed to use?

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,150 @@
...
+
+**Resource Mapping**
+This document sees us move from utilising the ResourceMapper to directly facilitating Resource to Domain Model mapping, but it does not directly address moving away from the mapper, or moving away from it when mapping from Domain Model => Resource when translating back to wire format on "the way out". This will be the focus of [another decision record]() TBA.
+
+**Domain Events**
+Domain event infrastructure helps us create rich, decoupled domain models that can evolve and scale in healthy ways. This PR does not introduce any new domain event infrastructure or considerations, although it is a topic we want to tackle soon after agreeing on the core domain approach and abstractions.
+
+## Overview
+
+[This PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834) works through an end-to-end example of refactoring the `Tenants` `PUT` and `POST` endpoints to use new commponents that align with the above principals.
```
andyinaus, line 37:
```suggestion
[This PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6834) works through an end-to-end example of refactoring the `Tenants` `PUT` and `POST` endpoints to use new components that align with the above principals.
```

Decisions/20200825-DomainModel.md
```diff
@@ -0,0 +1,150 @@
...
+
+To achieve this, we use:
+
+- ASP.NET Core's nullable reference type awareness, which will validate that information has been supplied for non-nullable fields during model binding.
+
+- Tiny Types, which we have a [custom serializer](https://github.com/OctopusDeploy/TinyTypes/blob/main/source/TinyTypes.Json/TinyTypeJsonConverter.cs) for, to ensure recieved information is in the expected format.
+
+- Fluent Validation, for when we have more complex validation needs that are still restricted to validating the structure of the information in a single model, [example](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Core/Validation/DaysPerWeekScheduledTriggerFilterValidator.cs).
+
+Once the model is bound and validated and is available to our controller to consume, we have verified the integrity of the data it contains, and that that data aligns with the type that is defined.
```
andyinaus, line 62:
```suggestion
Once the model is bound and validated and is available to our controller to consume, we have verified the integrity of the data it contains, and that data aligns with the type that is defined.
```

Decisions/20200825-DomainModel.md
andrewabest: I can see we have a number of `[Obsoleted]` methods already. Would this be a potential way forward with overloads?

I feel at some juncture we also need to revisit how we redefine our resource in Server and Octopus.Client - it might be good to do that when thinking about this. The resources in Server have a heap of unused methods on them, keeping resources in sync across the projects is painful at best, it isn't clear to me why they carry the same attributes in both server and client (assume there is a reason).

I've also removed the attributes from the `TenantWriteResource` - I'm not sure what the impact of this is, I understand that nevermore and the resource mapper was only one consumer of these attributes. @droyad do you know where else might be impacted if we move away from decorating our resources with these attributes?

Decisions/20200825-DomainModel.md
andrewabest: You steam a good ham @TomPeters. I do think that implicit conventions tend to be hard to discover / propogate in large codebases with lots of people working on them, but at the moment I'm happy to see how these document stores evolve and re-propose this particular approach if they get a bit unruly.

Decisions/20200825-DomainModel.md
droyad: A long time ago past me separated Octopus.Client out from the server code base. The resources went with it, and at the time I didn't remove a lot of the methods and attributes. There is a reason they are on both sides, just not a good one :).

If you are talking about the Writable/WriteableOnCreate/etc attributes they are mainly used to enforce what the user is allowed to set via ResourceMapping. Manually mapping removed the need for these (for that purpose). 

We were planning on using them for client and document generation however. The Create/Update model difference might mitigate that. It boils down to "Can a tool tell whether the field is writeable?" so that the correct client side code can be generated.

Decisions/20200825-DomainModel.md
droyad: > Each invalid request is now an expensive operation as most of the domain rules require accessing the data access layer to validate

In almost all cases the performance of a end-user data modification has not been a problem nor high priority for us. Modifications are relatively rare so they do not affect CPU/Perf/DB. 

From a UX perspective, the user rather wait a second or two for an update over having to attempt a save many times.

Decisions/20200825-DomainModel.md
droyad: I also don't think it supports some of the advanced serialization we do.

droyad (approved):
There seems to be consensus on this proposal so I'll give it the Approval stamp :).

For me the devil is in the detail. For me that's client code generation and API versioning, but this document doesn't make a decision on those aspects.

pawelpabich:
I really like that we rely on the compiler to do as much work as possible for us because nothing can provide feedback to engineers faster. 

I'm also glad to see that every new abstraction is required to earn its keep.

Decisions/20200825-DomainModel.md
pawelpabich: I can see that `IQuery` is gone now but the corresponding PR uses a simple `Func` as a way to perform IO.  It feels like there should be a section in this document that describes our recommended approach(es) to this problem.

andrewabest: I've added a paragraph for our preference on this now.

andrewabest:
The RFC period has ended, and all feedback responded to / incorporated into the document.

PR #21, opened 2020-08-26
DataMapping

andyinaus:
**Readable Version**: https://github.com/OctopusDeploy/Architecture/blob/data-mapping/Decisions/20200826-DataMapping.md

## Description
This is a newly proposed solution to replace the existing implicit `DataMapping` approach with an explicit approach. The new approach would not only provide better code navigation/discoverability/usage, debugging experience, and other benefits but offer us a good opportunity to address some of the tech debt(s) caused by the existing approach such as violation of single responsibility and incorrect use of `ILifetimeScope` etc, to improve our overall code quality. 

The proposed solution/implementation in this PR can be found at [this Spike PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6686) in `Octopus.Server` repository.

Moreover, a POC PR for migrating TaskDataMapping (`MixScopedResource`) can be found [here](https://github.com/OctopusDeploy/OctopusDeploy/pull/6913)

## Timeline
Aiming to have consensus on the approach presented by the end of sharpening (Friday 28 Aug) 🤞 .
We will leave this PR open for discussion for another two weeks (Friday 11th Sep)

- Comment on individual lines
- Comment on the PR. Consider using long-form replies detailing your position. Look for **FEEDBACK WANTED** annotations if you would like to understand the type of feedback we are directly looking for (although all feedback is welcome)
- Add 👍 comments or positions you support, it will be tallied when gauging the degree of consensus.

### DIFF
--- a/Decisions/20200826-DataMapping.md
+++ b/Decisions/20200826-DataMapping.md
@@ -1,0 +1,176 @@
+# Overview
+
+- **Subject**: Implementing A New Data Mapping Approach  
+- **Decision date**: [when was the decision made]
+
+# Executive Summary
+
+To replace the existing implicit data mapping approach with an explicit approach. The new proposed approach would not only provide better code navigation/discoverability/usage analysis, debugging experience, and other benefits but offer us a good opportunity to address some of the tech debt(s) caused by the existing approach such as violation of single responsibility and incorrect use of `ILifetimeScope` etc, to improve our overall code quality. 
+
+# Detail
+
+## Why do We wanna move away from implicit DataMapping (ResourceMapper)?
+
+- To eliminate any unseen/unpredictable bug(s) that could be caused by renaming properties of either Domain Model(s) or wire format (Resources)
+- To improve code navigation/discoverability/usage analysis as `Show Usage` cannot help us much e.g. Some required properties of wire format (Resources) have no usage at all, which may encourage developers to remove them. Moreover, some tool(s) such as static analysis of property usage may provide misleading outcome
+- To improve debugging experience e.g. Instead of stepping into some common/generic code implementation and looping through all the properties until reaching the specific property we are after, we could simply toggle a breakpoint at an executable line of code which corresponds to the property mapping
+- To remove the complexity of our implicit DataMapping such as how we make use of those special attributes (`Writeable`, `WriteableOnCreate` and `NotReadable`) to control Create, Read and Update operations, and how we construct those auto property mappings etc, so it's easier for developers to understand
+
+Apart from the above reasons, there are some existing problems/tech debts with our current implementation we would also like to avoid doing again. More details can be found [here](https://docs.google.com/document/d/1rb-dUOMsf7R6B5KjPbnAggJSn3p_rcqk3KokXWuAiQE/edit?usp=sharing).
+
+## Assumptions
+
+- We currently only employ this new `DataMapping` "on the way out", going from domain to wire format in 'the new world' of ASP.NET Core.
+(While the above assumption applies here, there isn't anything stopping us from using the proposed mapping shape elsewhere - it just isn't the focus of this PR)
+
+## Principals
+
+- We want to move toward writing explicit mapping code, and away from implicit/magic mapping (like [ResourceMapper](https://github.com/OctopusDeploy/OctopusDeploy/blob/de145f341f540247e8356e9db5452585e42e60df/source/Octopus.Server/Web/Mapping/ResourceMapper.cs#L17))
+- We want to make the new data mapping approach simple by only taking one single responsibility: mapping, so it's easier for any developers to understand
+- We want to increase the discoverability of each concrete mapping
+- We want to be able to migrate to explicit mappings easily
+- We want to make the new data mappings able to inject any dependencies they need
+- We want the new data mappings to comply with convention(s) like naming and inheritance etc
+- We want tests to be able to optionally supply stub mapping implementations or mock the implementation for tests of mapping dependents, such as ASP.NET Core controllers
+- We want each data mapping to represent a 1-1 relationship between an input and an output
+- We want each data mapping to be defined in one single class only
+- We want to improve the debugging experience
+
+## Exclusions
+
+- Where should this new `DataMapping` implementation live? For example, where might it lives so it can be referenced by `Octopus.Server` & `Sashimi`
+- Data projection from multi-sources
+
+## Brief Summary
+
+The new proposed approach is using explicit concrete mapping implementation to map between one source and one target. Each mapping derives from a common abstract class `DataMapping<TSource, TTarget>` for consistency. A Spike PR can be found [here](https://github.com/OctopusDeploy/OctopusDeploy/pull/6686), and please refer to `TestController` as a concrete example in the PR, it accepts a direct dependency of the concrete data mapping implementation `TestToTestResponseDataMapping` which aligns with most of the above principals apart from the convention and migration ones.
+
+To enforce our convention principal, we introduce the `DataMappingConventionFixture` to ensure every `DataMapping` implementation complies with the desired naming pattern and inheritance hierarchy.
+
+For our migration principal, it could be achieved differently depending on when the migration needs to happen. 
+More details will be discussed in the `Considerations` section.
+
+## Considerations
+
+To move away from the implicit `DataMapping` we have with `ResourceMapper`, we have considered the following explicit approaches:
+
+1. Use a static method for each mapping
+2. Introduce a new explicit data mapping approach, where each concrete mapping defines/implements the mapping between one source and one target. Each mapping derives from a common abstract class `DataMapping<TSource, TTarget>` for consistency (as demonstrated in the above PR)
+3. Another explicit data mapping approach mentioned at this [discussion thread](https://octopusdeploy.slack.com/archives/CTZT49JFJ/p1600048892014600) is instead of sharing the same shape (interface or generic methods) between each concrete mapping, we would like each concrete mapping to be able to pass any contextual information it needs for mapping e.g. `Map(VcsRunBook runBook, string gitref, string spaceId, string projectId, string runbookId)` etc. However, we are still uncertain about how the current VCS model and resource would be modelled as we currently have two different implementation(s) for VCS modelling e.g. `Project` and `VcsRunbook`
+
+We believe the second option is better than the others, because:
+
+1. Better consistency as each `DataMapping` has a similar shape so it's easier for developers to understand
+2. Better dependency graph as each `DataMapping` takes care of its dependencies
+3. Better testability as each `DataMapping` can be mocked/stubbed for testing
+
+### Migration Concern(s)
+
+Migration concern(s) could be addressed differently depending on when it needs to happen:
+1. Migrate independently in all Nancy responder(s) (before `Nancy Modules to ASP .Net Core Controllers` migration)
+2. Migrate along with or after `Nancy Modules to ASP .Net Core Controllers` migration
+
+#### Migrate independently in all Nancy responder(s)
+
+Unfortunately, we are currently unable to use concrete explicit data mapping in `Nancy` world, because there is a generic method `Collection<TResource>` in `CustomResponder`, which is trying to map each item from type `object` to a generic type `TResource`. 
+
+Thus, to migrate mappings independently, we need to implement a temporary solution. One of the possible solutions is introducing a new virtual method `MapModelToResource<TResource>(object model)` in `CustomResponder` along with an extra interface `IDataMapper` (containing a generic method signature `TTarget Map<TTarget>(object source)`) and its implementation. With these, responders may now override the mapping and use a new explicit `DataMapping` if desired. (Please refer to the above [PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6686) and [a POC PR for migrating TaskMapping](https://github.com/OctopusDeploy/OctopusDeploy/pull/6913) for more details)
+
+Alternatively, we could also make `IDataMapper` smarter so it knows when to fall back to use `ResourceMapper` when it cannot find the corresponding explicit data mapping between the given source and target type. 
+
+We can safely remove this temporary solution once we have successfully migrated all our `Nancy` responders.
+
+#### Migrate along with or after Nancy Modules to ASP .NET Core Controllers migration
+
+Migrating along with or after `Nancy Modules to ASP .NET Core Controllers` migration doesn't require a temporary solution, because we can still use the current `ResourceMapper` and safely migrate a specific mapping to the new explicit approach after converting all responder(s) in a nancy module to controllers completely. However, if we only wish to make use of ASP .NET controllers and still invoke the existing responders underneath each controller's action as the first step of the Nancy migration, we may still need to consider employing the above temporary solution.
+
+### Dependency Injection
+
+This is one of the existing problems we have with the `ResourceMapper`, as it takes direct dependency of `ILifeTimeScope` and passes the scope down to `ResourceMapping` for the following purpose(s):
+
+- Switching space context for `SpaceScopedRoutes`
+- Resolving any `IEnricher`(s) for after-mapping operation(s)
+
+For best practice, we should try our best to avoid taking `ILifeTimeScope` directly, because:
+
+1. It allows developers to manipulate/mess with the current `ILifeTimeScope` unnecessarily
+2. It creates a tightly-coupled relationship between `Autofac` and our codebase everywhere unnecessarily  
+
+To address this, we prefer the second or third option over the static method approach as they allow developers to inject any component(s)/service(s) required for mapping the data from source type to the desired target shape.
+
+As for the scope and lifetime of each `DataMapping`, we have considered the following two option(s):
+
+1. Register each `DataMapping` as `InstancePerLifetimeScope`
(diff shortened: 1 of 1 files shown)


Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+# Executive Summary
+
+TBA
+
+# Detail
+
+## Assumptions
+
+- We want to map from (Domain) `Model` to `Resource` only in 'the new world' of ASP.NET Core
```
johnsimons, line 14: I don't agree with this restriction.
I want to move us to manual mapping in Sashimi, so IMO we should add manual mapping to at least Sashimi

johnsimons: I would maybe even go one step further where we would support manual mapping in the old code to aid the migration to ASP.NET Controllers

andrewabest: I suspect what @andyinaus means here is that "on the way in", when we go from wire format to domain, we won't use mappers. We employ them "on the way out", going from domain to wire format. It might be worth mentioning that that while that assumption applies here, there isn't anything stopping us from using the proposed mapping shape elsewhere - it just isn't the focus of this PR.

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Objectives
+
+- We want to move away from auto mapping like our current `ResourceMapper`
```
andrewabest, line 18: I think it would be better to state the objective of what we want to move to

```suggestion
- We want to move toward writing explicit mapping code, and away from magic mapping (like `ResourceMapper`)
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+- We want to make the new data mapping approach simple by only taking one single responsibility: mapping, so it's easier for any developers to understand
+- We want to increse the discoverability of each concrete mapping
+- We want to be able to migrate our current auto data mapping to the new data mapping approach separately, so it can be migrated without any dependencies
```
andrewabest, line 21: Suggestion to make the intent of this point clearer:

```suggestion
- We want to be able to migrate to explicit mappings without needing to make significant changes to existing responder infrastructure
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+- We want to make the new data mappings able to inject any dependencies they need
+- We want the new data mappings to comply with convention(s) like naming and inheritance etc
+- We want ASP.NET Core controllers to be able to inject a mockable/testable concrete data mapping implementation
```
andrewabest, line 24:
```suggestion
- We want tests to be able to optionally supply stub mapping implementations for tests of mapping dependents, such as ASP.NET Core controllers
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+- We want each data mapping to be 1-1 relationship
```
andrewabest, line 25:
```suggestion
- We want each data mapping to represent a 1-1 relationship between an input and an output
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+- We want each data mapping to be defined in one single class only
+
+## Exclusions
+
+- Where should this new `DataMapping` implementation belong to? so it can be referenced by `Octopus.Server` & `Sashimi`
```
andrewabest, line 30:
```suggestion
- Where should this new `DataMapping` implementation live? For example, where might it live so it can be referenced by both `Octopus.Server` & `Sashimi`
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+- Data projection from multi-sources
+
+## Breif Summary
```
andrewabest, line 33:
```suggestion
## Brief Summary
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+[This PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/6686) demonstrates the new proposed approach to map data from one source to one target.
+Please refer to `TestController` as a concrete example, it injects the direct dependency of the concrete data mapping implementation `TestToTestResponseDataMapping` which align with most of the above principals apart from the convention and migration ones. More details will be discussed in the `Considerations` section.
```
andrewabest, line 36:
```suggestion
Please refer to `TestController` as a concrete example, it accepts a dependency of the concrete data mapping implementation `TestToTestResponseDataMapping` which aligns with most of the above principals apart from the convention and migration ones. More details will be discussed in the `Considerations` section.
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+To achieve the convention objective, we introduce a TestFixture `DataMappingConventionFixture` to ensure every `DataMapping` complies with the desired naming pattern and inheritance hierarchy.
```
andrewabest, line 38:
```suggestion
To enforce our convention objective, we introduce the `DataMappingConventionFixture` to ensure every `DataMapping` implementation complies with the desired naming pattern and inheritance hierarchy.
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+As for the migration one, a new virtual method `MapModelToResource<TResource>(object model)` is introduced in `CustomResponder` along with an extra interface `IDataMapper` (containing a method signature `TTarget Map<TTarget>(object source)`) and its implementation so each derived responder can override it once the new `DataMapping` is ready.
```
andrewabest, line 40:
```suggestion
For our migration objective, a new virtual method `MapModelToResource<TResource>(object model)` is introduced in `CustomResponder` along with an extra interface `IDataMapper` (containing a method signature `TTarget Map<TTarget>(object source)`) and its implementation. With these, responders may override the mapping and use a new explicit `DataMapping` if desired.
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+Note: We are unable to use concrete data mapping in `Nancy` world as the generic method `Collection<TResource>` in `CustomResponder` is trying to map each item from type `object` to a generic type `TResource`. 
+
+We can safely remove `IDataMapper` and its corresponding implementation once we have successfully migrate all our `Nancy` responders.
```
andrewabest, line 43:
```suggestion
We can safely remove `IDataMapper` and its corresponding implementation once we have successfully migrated all our `Nancy` responders.
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+To address the existing problem(s) we have with `ResourceMapper`, we have considered the following solution(s):
+
+1. Use static method for mapping
+2. Introduce a new `DataMapping` approach (as demonstrated in the above PR)
+
+We believe the second option is better than the first one, because:
+
+1. Better consisteny as each `DataMapping` has a similar shape so it's easier for developers to understand
+2. Beter dependency graph as each `DataMapping` takes care of its own dependencies
+3. Better testability as each `DataMapping` can be mocked for testing
```
andrewabest, line 60:
```suggestion
3. Better testability as each `DataMapping` can be stubbed for testing 
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+## Considerations
+
+To address the existing problem(s) we have with `ResourceMapper`, we have considered the following solution(s):
+
+1. Use static method for mapping
+2. Introduce a new `DataMapping` approach (as demonstrated in the above PR)
+
+We believe the second option is better than the first one, because:
+
+1. Better consisteny as each `DataMapping` has a similar shape so it's easier for developers to understand
```
andrewabest, line 58:
```suggestion
1. Better consistency as each `DataMapping` has a similar shape so it's easier for developers to understand
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+2. Beter dependency graph as each `DataMapping` takes care of its own dependencies
```
andrewabest, line 59:
```suggestion
2. Better dependency graph as each `DataMapping` takes care of its own dependencies
```

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Discoverability
+
+It's often very painful for developers to discover a file/class/interface in a very large codebase, especially when we use/inject an interface instead of the concrete type.
+In order to ease the pain and provide better maintenance/development experience to not just us but new developers, we would like to:
+
+- Always inject the concrete `DataMapping` type in our ASP.NET Core controller (which will be further enforced when we move `IDataMapping` interface into a separated library and hides its accessibiliy)
+- Use `ConventionTestFixture` to ensure every `DataMapping` complies with our naming pattern e.g. **{SourceTypeClassName}To{TargetTypeClassName}DataMapping**
+
+### Convetion(s)
```
andrewabest, line 88:
```suggestion
### Convention(s)
```

Decisions/20200826-DataMapping.md
johnsimons: I am not sure I entirely agree with this, IMO the manual mappings should be quite concrete so using the real thing would be my preferred option

andrewabest: I agree, that is why it is optional - we take a dependency on the concrete type deliberately, so unless the developer intervenes, you get the real deal in your test. But it is nice to have escape clauses if you need them, which we wouldn't get with options like static mapping methods.

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,144 @@
...
+
+This is another benefit given by the inheritance hierarchy. 
+To gain the testability, each `DataMapping` is forced to inherit from a generic type of `DataMapping<,>` which has a public abstract method `TTarget Map(TSource source)`. This abstract method will then make each concrete derived `DataMapping` type mockable.
+
+e.g.
+```
+var mapping = Substitute.For<TestToTestResponseDataMapping>();
+mapping.Map(Arg.Any<Test>()).Returns(new TestResponse("2", "2", "2"));
+```
+
```
andrewabest, line 135: To address @johnsimons's comment above, perhaps we could elaborate here that the desired approach would be the code under test uses the concrete mapper (which is why we take it as a dependency and don't depend on the abstraction), and that mocking would only be used for certain edge-case scenarios where it supports better testing.

Decisions/20200826-DataMapping.md
andyinaus: If we write E2E test, this may not help much. This may help with the following scenario:
In ControllerFixture - We ensure it's actually using the `DataMapping` to map the data
In MappingFixture - We ensure the `DataMapping` can map correctly in every possible scenario

matt-richardson:
I'm not entirely clear on the problem we're trying to solve here. I looked at the linked doc, and while it listed some issues and questions, I didn't come away compelled that we should switch to manual data mapping.
Consider adding a couple of sentences to the intro to give solid examples of where this has caused us to fall down and clarify the "why" before the "we want" part?

Decisions/20200826-DataMapping.md
```diff
@@ -0,0 +1,149 @@
...
+We believe the second option is better than the first one, because:
+
+1. Better consistency as each `DataMapping` has a similar shape so it's easier for developers to understand
+2. Better dependency graph as each `DataMapping` takes care of its own dependencies
+3. Better testability as each `DataMapping` can be mocked/stubbed for testing
+
+### Dependency Injection
+
+This is one of the existing problems we have with the `ResourceMapper`, as it takes direct dependency of `ILifeTimeScope` and pass the scope down to `ResourceMapping` for Space Context Switching and resolving `IEnricher`(s) for after-mapping operation(s).
```
TomPeters, line 65: **TL;DR I would like to see a POC of how we should change the way we generate our TaskResource Links**

I think there may be a blind spot that I would like to make you aware of, in case you weren't :)

The "hack" of taking a dependency on the LifetimeScope was done for a few reasons. I can think of 3 off the top of my head:
 1. Allows us to resolve our `IResourceEnricher`s and `IModelEnricher`s (and allow them to take any dependency they need) without polluting the `IResourceMappingContext` with things that are very specific to individual mappings. (there are plenty of other ways we could solve this specific problem, with your proposed solution being a good one :+1:)
 2. When doing data access, we have a stronger guarantee that the space scope of our data access layer (i.e. `IOctopusRelationalStore`) is scoped to the same space as the resource you are mapping. This means you can't step outside of the hard walls that we build up around spaces. See https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Server/Web/Mapping/ResourceMapping.cs#L78. 

	FYI, I think if we removed this bit of code, I can't find any usages of `IOctopusRelationalStore` in resource enrichers that I think would break, simply because we don't have any custom resource enrichers _yet_ for the kind of resources that are at risk (i.e. mixed scope resource enrichers, like for tasks/ScopedUserRoles/events/etc.). But it it still is a theoretical risk that could easily bite us if we aren't careful - I'd like us to think about this a bit more carefully first.

3. This is the main one and the reason for this design in the first place: The way we currently generate our routes that we populate on the Links on our resources depends on the space the resource is in. For example, the routes we put on the ProjectResource come from [here](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Server/Web/Routes/Api/ProjectRoutes.cs#L6), which you'll notice inherits from `SpaceScopedRoutes` and depends on an `ISpecificSpacePartition`. This ultimately means that we need an appropriate `SpacePartitionsDefinition` registered in the lifetime scope from which this type is resolved. This `SpacePartitionsDefinition` needs to match the spaceId from the model, hence the need to create a new one each time we map.

	Here's a motivating example that describes the problem better: I hit `/api/tasks?spaces=all&system=true` to get a list of tasks across all spaces, and also including system tasks. This could return tasks from `Spaces-1`, `Spaces-2` and `Spaces-3`.

	Within this responder, we get the list of tasks from the db, and then we individually map each of those tasks to `TaskResource`s.

	If a task resource happens to live within a space, then it contains a few extra links, like `Artifacts` and `Interruptions` ([here](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Server/Web/Mapping/ResourceMappings.cs#L410)). These are links that are specific to spaces, so need to contain a space id (like `/api/Spaces-1/artifacts/...`. In fact, this space id must be the same as the space id for that task!

	For each individual task, we create a new lifetime scope that has its space context set to the space for that task, so that when we resolve the links for that task they get configured with the right spaceId.

	As it stands, I don't think your current proposal will work for task resources links because of this problem.

	FWIW, the problem was _much_ worse at an earlier stage in the design of Spaces (due to some different decisions we had made about what our links would look like). I just had a look and as far as I can tell, the TaskResource links might be the _only_ place affected by this problem! So it might not actually be a big deal.

andyinaus: Thanks for your valuable comment 💯 😄 

For the 3rd point, wondering if making `spaceId` substitutable in `SpaceScopedRoutes` would help or not.
IMO, this sounds more like an issue with `AllRoutes` rather than `DataMapping` itself, becuase `ServerTask` or anything implements `IHaveSpace` interface is already aware of its own `SpaceId`, we shouldn't try to switch the space context and start a new child `ILifetimeScope` just so the `ISpecificSpacePartition` dependency in `SpaceScopedRoutes` can get the correct `SpaceId`. This solution seems too powerful for a problem like this to me. (There may be some important reason(s) why this implementation was introduced before that I'm not fully aware of. If I sound stupid here, please forgive my ignorance)

And if we really need to switch the space context using the current approach for our second or even the third concern here, the worst case is ... we still have to employ the existing space switching implementation logic in the new data mapping design until we are confident that we no longer need to switch between spaces within a single lifetime scope anymore. 
 
I would love to deliver a POC for `ServerTask` mapping, it will help me understand more about our space design here. 😄

I really appreciate your time and comment(s). 

I always learn a lot from your comment(s) 🙇‍♂️ 🙇‍♂️ 🙇‍♂️ 🙇‍♂️ 😄

TomPeters: >There may be some important reason(s) why this implementation was introduced before that I'm not fully aware of.

I don't think you're missing anything. We just have a lot of routes and this was the cheapest way to hook it all up in a safe/reliable way. Other options were much more expensive.

andyinaus:
> I'm not entirely clear on the problem we're trying to solve here. I looked at the linked doc, and while it listed some issues and questions, I didn't come away compelled that we should switch to manual data mapping.
> Consider adding a couple of sentences to the intro to give solid examples of where this has caused us to fall down and clarify the "why" before the "we want" part?

Apologies for the confusion. I will try to outline the "why" part clear here in this PR

Decisions/20200826-DataMapping.md
pawelpabich: What if only Integration/E2E tested the happy path in the controllers? From what I've seen, testing at the ASP.NET MVC or Nancy Responder level requires you to mock half of the world. A real-life example should help make the discussion here more concrete.

andyinaus:
Going to close this one. A new data mapping [PR](https://github.com/OctopusDeploy/Architecture/pull/27) is created

PR #22, opened 2020-09-11
External Tool Dependency Versions decision document

andrewabest:
Readable Version: https://github.com/OctopusDeploy/Architecture/blob/dependency-versions/Decisions/20200911-ExternalToolDependencyVersions.md

Timeline: Input will be recieved over the next week, closing Friday 18/9.

Description
This PR summarises our stance on managing dependency versions for external tools - what version should we depend on, what impacts that decision has on us, and what our overall stance is for dependencies on external tools.

How to engage with this PR:

Add 🤔 to the PR to indicate you want to be consulted before it is close as you are still considering it or are formulating a reply
Comment on individual lines
Comment on the PR. Consider using long form replies detailing your position. Look for FEEDBACK WANTED annotations if you would like to understand the type of feedback we are directly looking for (although all feedback is welcome)
Add 👍 comments or positions you support, it will be tallied when gauging degree of consensus.
Submit a PR against this branch (and link in the comments) to make additions or propose changes
Submit a competing proposal in a new branch and PR (and link in the comments)

### DIFF
--- a/OctopusServer/ExternalToolDependencyVersions.md
+++ b/OctopusServer/ExternalToolDependencyVersions.md
@@ -1,0 +1,68 @@
+# Overview
+
+- **Subject**: Deciding What Version To Take Of An External Tool Dependency
+
+# Executive Summary
+
+In Octopus, we package external tools in various pieces of our software that allow Octopus to interact with and deploy to external services. Examples of these tools include the Terraform CLI and Azure CLI.
+
+We will pin all dependencies on external tools to a single static version, as this provides the most stable and reliable way for us to manage these dependencies over time.
+
+# Detail
+
+You can see what tools we package and embed in Octopus Server [on GitHub](https://github.com/octopusdeploy/?q=Octopus.Dependencies.&type=&language=).
+
+Another example of where we package and embed external tools is in our [Dynamic Worker images](https://github.com/OctopusDeploy/DynamicWorkerVmImages). We load both our packaged dependencies onto dynamic workers in their [installer scripts](https://github.com/OctopusDeploy/DynamicWorkerVmImages/blob/master/Win2019/scripts/install-tool-packages.ps1), and other external tools in other installer scripts, such as [kubectl](https://github.com/OctopusDeploy/DynamicWorkerVmImages/blob/master/Win2019/scripts/install-kubectl.ps1).
+
+Bugs have arisen that are the direct result of us taking the 'latest' version of an external tool when building and packaging our software ([example 1](https://trello.com/c/uufTKTnY/3699-breaking-terraform-upgrade-has-been-deployed-across-octopus-cloud), [example 2](https://trello.com/c/NxqZvQ4U/3724-dynamic-worker-images-are-pre-loading-the-incorrect-version-of-some-tooling-packages)).
+
+## Considerations
+
+### Taking Latest At Build Time
+
+Taking latest means to query some external feed as to what the latest available version of a tool is, and to use that regardless of what version number it is.
+
+Taking the latest tool as a dependency ensures we have the latest patches and security updates for the tools we embed in our products.
+
+If we take the latest at build time, it can be difficult to predict if this would have a negative impact on our software that depends on it. It can also be difficult to predict if it would have a negative impact on our customer's deployment processes, which also may utilize the packaged dependency.
+
+Whilst we have a level of automated testing that exercises our software and its interactions with these tool dependencies, they could never be extensive enough to predict all behaviour changes that could be introduced in newer versions of these tools.
+
+### Pinning Dependencies
+
+Pinning dependencies means to hard-code the version of tool we embed in our product, and to request this version each time our tool dependency packages are re-packaged and build.
+
+Pinning dependencies means we do not get the latest patches and security updates for the tools we embed in our products. We only get these updates when we deliberately increase the version of tool we depend on. This can be challenging, as our efforts are often concentrated elsewhere, and dependencies tend to be left locked to certain versions for long periods of time.
+
+By pinning our tool dependencies, we ensure that our own code, and our customers deployment scripts, will always run reliably as our software evolves and is rebuilt over time. We won't "accidentally" break our own software, or our customer's scripts, by rebuilding a dependency package, or revving the version of a dependency package depended on in Server, Dynamic Workers or other locations.
+
+### Tools Vs Libraries
+
+Drawing a distinction between tool versions and library versions is important. People other than us depend on tool versions, so our test suites are never going to have enough coverage. We're the only ones who depend directly on libraries, so it's a not-unreasonable goal to have good enough test coverage for the latter. In this document we are addressing **tools**, not libraries.
+
+### Version Ranges
+
+Some software package managers allow version ranges to be specified, which allows you to 'safely' increase your version dependencies - for example you can pin your major and minor version, but allow for any new patch versions to be utilised.
+
+This works if the world runs on SemVer. But it does not, as Terraform demonstrated in its `0.11` => `0.12` version bump that contains a large amount of breaking changes.
+
+For us, being explicit about a single version for each tool we depend on makes our software easy to reason about and support.
+
+### Handling Transient Dependencies
+
+For us, often the packaged tool is a transient dependency. An example of this is:
+
+`Octopus.Server` depends on `Sashimi.Terraform` which depends on `Octopus.Dependencies.TerraformCLI` which packages the Terraform CLI.
+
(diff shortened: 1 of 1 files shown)


Decisions/20200911-ExternalDependencyVersions.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Taking Latest At Build Time
+
+Taking latest means to query some external feed as to what the latest available version of something is, and to use that regardless of what version number it is.
+
+Taking the latest dependency ensures we have the latest patches and security updates for the tools we embed in our products.
+
+If we take the latest dependency at build time, it can be difficult to predict if this would have a negative impact on our software that depends on it. It can also be difficult to predict if it would have a negative impact on our customer's deployment processes, which also may utilize the packaged dependency.
+
+Whilst we have a level of automated testing that exercises our software and its interactions with these dependencies, they could never be extensive enough to predict all behaviour changes that could be introduced in newer versions of these dependencies.
```
uglybugger, line 30: This one's going to be interesting. I think it's a "horses for courses" decision but that we need some good principles around it.

For instance, we deliberately take the latest 3.1 floating tag for the .NET Core + Framework Docker images. We deliberately do this because the risk (in this case) is low and the reward is high.

Contrast that with taking the latest in Terraform, and it's a very different risk/reward profile.

Decisions/20200911-ExternalDependencyVersions.md
```diff
@@ -0,0 +1,69 @@
...
+
+If we allow our build process to take the latest version of a dependency, we cannot appropriately apply SemVer to our own packages, and therefore cannot predict when we are updating dependencies in upstream packages, what the impact of those updates would be.
+
+By pinning our dependencies, we can rev our SemVer's appropriately when we deliberately increase our hard-coded dependency version.
+
+For example: when we update `Octopus.Dependencies.TerraformCLI` to Terraform `0.12`, we will rev the major version to `2.x`. `Sashimi.Terraform` would need to consider that breaking change and see if it could take the new dependency without introducing breaking changes itself, and so on up the dependency change. By pinning our dependencies, we can be explicit about SemVer.
+
+## Decision
+
+We will pin all dependencies on external tools to a single static version, as this provides the most stable and reliable way for us to manage these dependencies over time.
```
uglybugger, line 62: Can we clarify this to say "We pin a single, static version" but also that "We update the pinned versions automatically, with no human intervention, after appropriate test suites have passed"? This allows for us to enable things like Dependabot + Bulldozer when we're sufficiently confident in our test suites.

Decisions/20200911-ExternalDependencyVersions.md
uglybugger: Another example of "horses for courses" within the `build-base-dotnet` image space: we use the `mcr.microsoft.com/dotnet/core/sdk:3.1` floating tag because we want the security updates, bug fixes, framework version updates etc., but in that same `Dockerfile` we deliberately pin the versions of GitVersion, Cake, NUKE etc and have humans version-bump them as required.

Decisions/20200911-ExternalDependencyVersions.md
```diff
@@ -0,0 +1,69 @@
...
+
+Pinning dependencies means we do not get the latest patches and security updates for the tools we embed in our products. We only get these updates when we deliberately increase the version we depend on. This can be challenging, as our efforts are often concentrated elsewhere, and dependencies tend to be left locked to certain versions for long periods of time.
+
+By pinning our dependencies, we ensure that our own code, and our customers deployment scripts, will always run reliably as our software evolves and is rebuilt over time. We won't "accidentally" break our own software, or our customer's scripts, by rebuilding a dependency package, or revving the version of a dependency package depended on in Server, Dynamic Workers or other locations.
+
+### Version Ranges
+
+Some software package managers allow version ranges to be specified, which allows you to 'safely' increase your version dependencies - for example you can pin your major and minor version, but allow for any new patch versions to be utilised.
+
+This works if the world runs on SemVer. But it does not, as Terraform demonstrated in its `0.11` => `0.12` version bump that contains a large amount of breaking changes.
```
droyad, line 44: Technically this `0.11` -> `0.12` is a breaking change (https://semver.org/#spec-item-4)

droyad:
Right now we have no intention of ever updating any of the bundled tools again (Terraform, AzurePS, AWS CLI etc). We are going to deprecate them in favour of the container tools.

I think we should take the stance of "don't touch the version".

Decisions/20200911-ExternalDependencyVersions.md
droyad: On all the tools that this applies I think we are so far behind now it's not worth considering changing it to floating versions.

Decisions/20200911-ExternalDependencyVersions.md
uglybugger: Well, technically, since they're still on version 0, _anything_ may change at any time. But that's ridiculous :p

Decisions/20200911-ExternalDependencyVersions.md
andrewabest: We can't state that at the moment - the terraform bump happened automatically, the test suits passed, and it broke the world.

It may be worth making this document more explicitly focussed on tools. The breakages that spurred this document were all tool dependencies - executables and the like. They aren't retrieved from package managers, they tend to be downloaded from somewhere. There may not be a nice way to intelligently bump some of them (i.e. for patch updates) - it is either "get latest" or "get specific version".

andrewabest:
@droyad that is good context - do you think you could contribute to the overview somewhere a paragraph summarising our future intent for these tools?

Decisions/20200911-ExternalDependencyVersions.md
uglybugger: Agreed - I think drawing a distinction between tool versions and library versions is important. People other than us depend on tool versions, so our test suites are never going to have enough coverage. We're the only ones who depend directly on libraries, so it's a not-unreasonable goal to have good enough test coverage for the latter.

andrewabest: I've updated the document to make it clearer we are talking about dependencies on tools, not libraries.

Decisions/20200911-ExternalDependencyVersions.md
andrewabest: Haha sure, I suppose it does.

michaelnoonan:
This seems reasonable from my perspective. :+1:

andrewabest:
@droyad are you happy for this to be merged? I've incorporated all of the feedback recieved so far.

Decisions/20200911-ExternalDependencyVersions.md
pawelpabich: > Agreed - I think drawing a distinction between tool versions and library versions is important. People other than us depend on tool versions, so our test suites are never going to have enough coverage. We're the only ones who depend directly on libraries, so it's a not-unreasonable goal to have good enough test coverage for the latter.

I like this distinction because it's based on how we use dependencies.  

That being said, I think that talking about `tools`, `libraries`, etc muddies the waters here a bit.  E.g. `dotnet` looks like a tool but we are not pining its version in our build scripts. [ `material-ui` ](https://github.com/OctopusDeploy/OctopusDeploy/blob/5bcfc53b28e31a3452f1277706b3c5747f623073/newportal/package.json#L204)is a library but we would be crazy :) to depend on a version range based on our past experiences. 

What about replacing `tools`, `libraries`, `executables` with `dependencies`? Then, if a dependency is exposed directly to our customers (e.g. terrafrom.exe) we will pin its version. If a dependency is used only by us then, I believe,  the default position still should be to pin its version.  Relying on a range of versions would be recommended when the benefits significantly outweigh the risks. E.g. Microsoft applies mostly only security-related patches to their LTS releases and makes an effort not to introduce any breaking changes.

andrewabest: This will make the document a lot more "general purpose". At the moment it is focussed on addressing our dependency on external tools - the detail, the considerations are all focussed on tools. 

The reason for this is there was ambiguity in the previous revision which talked about "dependencies", where @uglybugger asked us to mention that ideally we want to be able to automatically update our dependencies on green builds. We do not want to do that for tools. 

If I went and changed all references to "tools" to "dependencies", I feel like these subtleties would still need to be addressed.

We would need to be specific about:

- Framework dependencies like `dotnet` - the core building blocks we build our software on
- Library dependencies - smaller lego bricks we use within our software, often open source
- Tool dependencies - things we integrate with to provide a way of driving external services
- Other things?

I'd like a little more guidance / input before reformatting the document to address these things, as I feel like if I did it it might still miss the mark.

PR #26, opened 2020-10-19
Add guidelines for writing migration scripts

flin-8:
# Description

It's not clear when we need to write migration scripts. This is a source of confusing bugs given the pitfalls involved.
This document hopes to make it easy to make a decision.

Feel free to edit/suggest.

### DIFF
--- a/OctopusServer/Migrations.md
+++ b/OctopusServer/Migrations.md
@@ -1,0 +1,31 @@
+# Migration Scripts
+
+## Overview
+
+When making changes that affect the state/shape of persisted objects, we should always write a migration script to upgrade existing objects in the database. This is the case even if only pre-release versions of Octopus are affected, as our internal instances as well as Octopus Cloud may have been exposed to that version.
+
+The other benefit is reduced ambiguity, where we don’t rely on:
+- Serialization behaviour to remove old properties
+- Mainline code to add default values for new properties.
+
+## Problem
+
+Currently it's not clear when we need to write migration scripts when the shape of a persisted object changes, since it’s possible to lean on mainline code to fill in default values for new properties, and serialization to remove old properties.
+
+There are also cases where a bug prevented a property from being filled out, should we go to the trouble of writing a migration script if we are just fixing that bug?
+
+## Proposed Solution
+
+If objects created/manipulated before and after the change look different when persisted, then write a migration script.
+
+This ensures that:
+- Objects created/manipulated before and after the change look identical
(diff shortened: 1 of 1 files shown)


MarkSiedle (approved):
Love it + 100% agree. Adding various "defaults" or "conditional fill-ins" in our mainline code (because something may or may not exist) is horrible for future-us to have to reason about. We've still got various hacks in our frontend to protect against bad data/missing defaults etc.  that were very difficult to work through in the step-editor revamp for example.

You mentioned this was pretty raw, but I think this sums up the problem/solution/direction really well. It's a great foundation that we can iterate on in terms of guidance.

This is not a blocker, but do we have any recent PRs (from QRF or otherwise) that have resolved one of these issues? I had a quick look through the closed-PRs list, but none of the titles jumped out at me. If you know of any, would it be worth including an example PR that addresses/solves one of these problems (showing our recommended approach for DB migrations like these)?

LGTM

flin-8:
> Love it + 100% agree. Adding various "defaults" or "conditional fill-ins" in our mainline code (because something may or may not exist) is horrible for future-us to have to reason about. We've still got various hacks in our frontend to protect against bad data/missing defaults etc. that were very difficult to work through in the step-editor revamp for example.
> 
> You mentioned this was pretty raw, but I think this sums up the problem/solution/direction really well. It's a great foundation that we can iterate on in terms of guidance.
> 
> This is not a blocker, but do we have any recent PRs (from QRF or otherwise) that have resolved one of these issues? I had a quick look through the closed-PRs list, but none of the titles jumped out at me. If you know of any, would it be worth including an example PR that addresses/solves one of these problems (showing our recommended approach for DB migrations like these)?
> 
> LGTM

I've added an example of a fix to the document (https://trello.com/c/Z5SnrNid/3764-variable-substitution-broken-for-steps-based-on-a-template)

flin-8:
@NickJosevski did you want to add anything here?

NickJosevski:
Looks good thanks for pinging me, this is an area I care a lot about in server as often I had to fix it.

If you need more examples, look back at old migration scripts, will be faster to spot repairs and the associated PRs + Issues

Good stuff mentioning:
 * pre-release versions
 * Serialization behaviour
 * Mainline code to add default

The mainline defaults/repairs - I'd call this out as something that **must be avoided and cleaned up if spotted.**

 > Remove any mainline patching code - as it only delays the bug hitting us. The more time that passes the more chance for the data to drift even further (subsequent patches may miss it if it's not recognisable).

NickJosevski (approved):
😄

PR #27, opened 2020-10-21
Add data mapping ADR

andyinaus:
**Readable Version**: https://github.com/OctopusDeploy/Architecture/blob/data-mapping-decision-record/Decisions/20201014-DataMapping.md

## Description
This ADR is to replace the existing implicit data mapping approach with an explicit approach. The new approach would 
allow developers to choose the most suitable/reasonable way to define/implement a data mapping explicitly depending on different circumstances.

## Timeline
Aiming to have consensus on the approach presented by the end of sharpening (Friday 28 Aug) 🤞 .
We will leave this PR open for discussion for another two weeks (Friday 11th Sep)

- Comment on individual lines
- Comment on the PR. Consider using long-form replies detailing your position.
- Add 👍 comments or positions you support, it will be tallied when gauging the degree of consensus.

### DIFF
--- a/Decisions/20201014-DataMapping.md
+++ b/Decisions/20201014-DataMapping.md
@@ -1,0 +1,26 @@
+# Overview
+
+- **Subject**: [explain what decision you are making]
+- **Decision date**: [when was the decision made]
+
+# Executive Summary
+
+Each explicit data mapping can be defined/implemented variously depending on whether testing, reusability and dependencies are required or not.
+
+# Detail
+
+## Considerations
+
+Captured in this document: https://docs.google.com/document/d/1C-4DRUEaER_zWmecM5lECkZAHOie3mlaMWPZFJnfg-k/edit?usp=sharing
+
+## Decision
+
+If a data mapping requires any dependencies, use a separate class to resolve its dependencies and always use/inject a concrete data mapping type for better discoverability, e.g. [MapToVcsRunbookResources](https://github.com/OctopusDeploy/OctopusDeploy/blob/4702ef0677895d7d41f22af337e0d66f736f787c/source/Octopus.Server/Web/Mapping/MapToVcsRunbookResources.cs#L12). If dependencies are not required, use a private method (e.g. [MapToRunbookDomainModel](https://github.com/OctopusDeploy/OctopusDeploy/compare/spike-tenant-domain-model...spike-tenant-domain-model-runbooks#diff-a5c39abbfe2fc86ceb6451588fa356b4a2d73ed1504d87f1a230686ee6c5bd6eR89)) when both testing (in isolation) and reusability of a data mapping are not necessary, otherwise, use a static method.
+
+## Data Sources
+
+Source listed below might provide additional context but keep in mind that they can disappear at any time.
+
+- [Different Explicit Data Mapping Solutions](https://docs.google.com/document/d/1C-4DRUEaER_zWmecM5lECkZAHOie3mlaMWPZFJnfg-k/edit?usp=sharing)
+- [Project Config as Code Feedback](https://octopusdeploy.slack.com/archives/C01AJE4K3T2/p1602486226148900)
+- [Initial Discussion PR](https://github.com/OctopusDeploy/Architecture/pull/21)


andrewabest:
Please ensure this PR conforms with the new guidance on structure in https://github.com/OctopusDeploy/Architecture/blob/master/README.md

pawelpabich:
@andrewabest, I'm still  getting used to the new structure.  Assuming it is an ADR, which it seems to be, it's not clear to me what needs to be changed? Move it under `Decisions\OctopusServer` folder?

pawelpabich (approved):
Content looks good to me. I will let @andrewabest comment on the structure.

andrewabest:
The file should be moved to the `OctopusServer` folder, and be renamed to `DataMapping.md` as we have removed date stamping from files.

Decisions/20201014-DataMapping.md
```diff
@@ -0,0 +1,26 @@
+# Overview
+
+- **Subject**: [explain what decision you are making]
+- **Decision date**: [when was the decision made]
+
+# Executive Summary
+
+Each explicit data mapping can be defined/implemented variously depending on whether testing, reusability and dependencies are required or not.
```
andrewabest, line 8: This summary does not tell me a lot - what does `variously` mean? We should be summarizing the decision / specifics, and this is a very general statement.

Decisions/20201014-DataMapping.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Detail
+
+## Considerations
+
+Captured in this document: https://docs.google.com/document/d/1C-4DRUEaER_zWmecM5lECkZAHOie3mlaMWPZFJnfg-k/edit?usp=sharing
+
+## Decision
+
+If a data mapping requires any dependencies, use a separate class to resolve its dependencies and always use/inject a concrete data mapping type for better discoverability, e.g. [MapToVcsRunbookResources](https://github.com/OctopusDeploy/OctopusDeploy/blob/4702ef0677895d7d41f22af337e0d66f736f787c/source/Octopus.Server/Web/Mapping/MapToVcsRunbookResources.cs#L12). If dependencies are not required, use a private method (e.g. [MapToRunbookDomainModel](https://github.com/OctopusDeploy/OctopusDeploy/compare/spike-tenant-domain-model...spike-tenant-domain-model-runbooks#diff-a5c39abbfe2fc86ceb6451588fa356b4a2d73ed1504d87f1a230686ee6c5bd6eR89)) when both testing (in isolation) and reusability of a data mapping are not necessary, otherwise, use a static method.
```
andrewabest, line 18:
```suggestion
If a data mapping requires any dependencies, use a separate class to resolve its dependencies and always use/inject a concrete data mapping type for better discoverability, e.g. [MapToVcsRunbookResources](https://github.com/OctopusDeploy/OctopusDeploy/blob/4702ef0677895d7d41f22af337e0d66f736f787c/source/Octopus.Server/Web/Mapping/MapToVcsRunbookResources.cs#L12). 

If dependencies are not required, use a private method (e.g. [MapToRunbookDomainModel](https://github.com/OctopusDeploy/OctopusDeploy/compare/spike-tenant-domain-model...spike-tenant-domain-model-runbooks#diff-a5c39abbfe2fc86ceb6451588fa356b4a2d73ed1504d87f1a230686ee6c5bd6eR89)) when testing of the mapping itself, and reuse of the mapping, is not required. Otherwise, use a static method.
```

Decisions/20201014-DataMapping.md
andrewabest, line 8:
```suggestion
Data mappings that required dependencies will be defined in classes that can be resolved and injected with those dependencies. Those that do not will be defined in private methods if testing and reuse are not required, or static methods if they are.
```

Decisions/20201014-DataMapping.md
```diff
@@ -0,0 +1,26 @@
+# Overview
+
+- **Subject**: [explain what decision you are making]
```
andrewabest, line 3: Let's make sure these are filled in before merging.

pawelpabich:
> The file should be moved to the OctopusServer folder, and be renamed to DataMapping.md as we have removed date stamping from files.

Looking at https://github.com/OctopusDeploy/Architecture/tree/master/Decisions this is an ADR. It's not clear to me why https://github.com/OctopusDeploy/Architecture/tree/master/Decisions is empty. I will move this conversation to #topic-arch

PR #30, opened 2020-11-12
Name and Id Considerations

MarkSiedle:
[Readable version](https://github.com/OctopusDeploy/Architecture/blob/mark/cac/nameorids/OctopusServer/ConfigAsCode/NamesOrIds.md)

This cycle, we had a chance to review our current approach to name/ids with a fresh perspective. We had some new eyes on the approach, previous questions being re-raised and some joint concerns have been raised. The most notable concern is: How far the current approach is spreading. But also, there’s concerns around the user-experience for consumers when we deliver this feature.

This document is an attempt to capture all of this information to make sure we’re still heading in the right direction.

### DIFF
--- a/OctopusServer/ConfigAsCode/NamesOrIds.md
+++ b/OctopusServer/ConfigAsCode/NamesOrIds.md
@@ -1,8 +1,8 @@
 # Name and Id Considerations
 
-This cycle, we had a chance to review our current approach to name/ids with a fresh perspective. We had some new eyes on the approach, previous questions being re-raised and some joint concerns have been raised. The most notable concern is: How far the current approach is spreading. But also, there’s concerns around the user-experience for consumers when we deliver this feature.
+This cycle, we had a chance to review our current approach to name/ids with a fresh perspective. We had some new eyes on the approach, previous questions being re-raised and some joint concerns have been raised. The most notable concern is: **How far the current approach is spreading**. But also, there's concerns around the user-experience for consumers when we deliver this feature.
 
-This document is an attempt to capture all of this information to make sure we’re still heading in the right direction.
+This document is an attempt to capture all of this information to make sure we're still heading in the right direction.
 
 Note: This was converted from a [Google Doc discussion](https://docs.google.com/document/d/1J_Vjw1pH5t9ptUI2F6NLyAlSn7WvGBh9LeL1HrUUPdQ/edit).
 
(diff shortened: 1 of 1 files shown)


OctopusServer/ConfigAsCode/NamesOrIds.md
```diff
@@ -0,0 +1,99 @@
...
+### Advantages
+- Names in OCL are human-friendly
+- Consumers of the API generally work in terms of names (be it from the command line, or a script or Hosted configuration). Using names means they don’t have to look up IDs.
+- We don’t pay any conversion-cost of Name-Ids until we’re ready to snapshot
+- You can provide a Name that doesn’t yet exist (but you know _will_ exist when you go to snapshot things)
+
+### Disadvantages
+- Any software-clients that call our API need to be updated to send Names. E.g.
+  - our frontend needs switching logic everywhere
+  - various octo.exe tooling may need to be updated accordingly as well
```
droyad, line 31: We want to move the logic in octo.exe into the server. When we do we'd likely want to expose things like create-release as an API that maps the command line parameters into a JSON object to post. The client won't need to map Names  to IDs.

johnsimons: Maybe this is the new command/action based API we have been talking about

OctopusServer/ConfigAsCode/NamesOrIds.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Advantages
+- Names in OCL are human-friendly
+- Consumers of the API generally work in terms of names (be it from the command line, or a script or Hosted configuration). Using names means they don’t have to look up IDs.
+- We don’t pay any conversion-cost of Name-Ids until we’re ready to snapshot
+- You can provide a Name that doesn’t yet exist (but you know _will_ exist when you go to snapshot things)
+
+### Disadvantages
+- Any software-clients that call our API need to be updated to send Names. E.g.
+  - our frontend needs switching logic everywhere
```
droyad, line 30: The end plan is that the old API goes away

OctopusServer/ConfigAsCode/NamesOrIds.md
```diff
@@ -0,0 +1,99 @@
...
+- As engineers, there’s mounting complexity we’re adding here
+  - Switching logic everywhere / harder to reason about things
+  - Additional test coverage needed everywhere
+  - “It’s spreading like a virus” is something I’ve heard a couple of times now
+- As a customer, I can’t take my existing JSON or scripts and use them for CaC projects
+  - “Wait, I need to figure out the name for all the IDs I have in my scripts?” :sadpanda:
+- What is the benefit of supporting Names in our API, for consumers?
+  - Arguably, the common pattern is GET, then a POST/PUT, so you have the whole record available to you. Why support Name?
+- The primary feature is to support config-as-code, not Names support in the API
+  - Are we coupling Name support in the API to the config-as-code feature prematurely? And making all of this more complex that it needs to be?
```
droyad, line 43: I don't think so. It's a major part of the design of the feature. It is also risky, front loading it will (has) reduce that.

MarkSiedle (commented):
This summarises the decisions made to this point. Merging, so future-us can continue the discussion.

PR #31, opened 2020-11-12
Added some more advantages to Current Approach

droyad:
Amends #30


OctopusServer/ConfigAsCode/NamesOrIds.md
```diff
@@ -24,6 +24,9 @@ Discussion: [See Slack thread](https://octopusdeploy.slack.com/archives/C01AJE4K
 - Consumers of the API generally work in terms of names (be it from the command line, or a script or Hosted configuration). Using names means they don’t have to look up IDs.
 - We don’t pay any conversion-cost of Name-Ids until we’re ready to snapshot
 - You can provide a Name that doesn’t yet exist (but you know _will_ exist when you go to snapshot things)
+- For EAP we want to see what a name based API could look like and our customers reactions to that
+- We are already well down this road and have it working
+- If we were designing the API today, this might be way we would have chosen to do it
```
tothegills, line 29: I think we would go all names though. The awkward thing that I'm running into is that we refer to resources by name in the deployment process but need to retrieve resources from the API by id.

MarkSiedle:
Merging this for us moving forward to learn from.

PR #35, opened 2021-01-22
Branch selector location

MarkSiedle:
Every cycle, we tend to ask _"Why is the branch selector not just on the sidebar so I know which branch I'm working in?"_.

Because ... reasons :) Let's see if we're convinced.

### DIFF
--- a/OctopusServer/ConfigAsCode/BranchSelectorLocation.md
+++ b/OctopusServer/ConfigAsCode/BranchSelectorLocation.md
@@ -24,10 +24,10 @@ When you navigate into a project and are in the mindset of creating a release, w
 
 ### Advantages
 
-- it's a single and consistent place for the user to learn where branch selecting/switching occurs
-- it removes the need for the GitRef field in the Create Release process (one less field to think about)
+- it's a single and consistent place for the user to learn where branch selecting/switching occurs = less to think about
+- it would simplify the GitRef field in the Create Release process (turns it into a read-only field where we tell the user which GitRef branch they're currently on, rather than a selection) = less to test
 - it's less maintenance in our codebase (rather than scattering the branch selector throughout various components) = less to test and arguably less risk
 
 ### Disadvantages
 
-- the user may be confused about which things are version controlled in the project. The current design is very clear on which things are version-controlled
+- the user may be confused about which things are version controlled in the project
--- a/OctopusServer/ConfigAsCode/BranchSelectorLocation.md
+++ b/OctopusServer/ConfigAsCode/BranchSelectorLocation.md
@@ -1,0 +1,33 @@
+# Branch Selector Location
+
+Every cycle, we tend to ask _"Why is the branch selector not just on the sidebar so I know which branch I'm working in?"_.
+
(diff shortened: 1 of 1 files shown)


OctopusServer/ConfigAsCode/BranchSelectorLocation.md
```diff
@@ -0,0 +1,41 @@
...
+
+When you load a project (E.g. just going to the Overview project page), there are various things that still need to know which branch you're working on.
+
+For example:
+- Various sidebar links need to embed the branch into their routes
+- There are network requests (to the deployment process API) to decide whether we even show the `Create Release` button. So knowing which branch you're on is important to the _whole_ screen, not just select paper layouts where version-control applies.
+
+Moving the Branch Selector up into a central place may simplify a lot of things and is actually consistent with what the user is seeing as a whole.
+
+### Advantages
```
bonnietsang-octopus, line 33: - The branch selector is a global concept within a project, it would be nice to have it in a consistent place. Having the selector on individual pages make it looks like an independent idea. Let’s not chop the concept into pieces; it’s good for our user to see the project as a whole.
- Having the selector on mutiple pages may give the impression that they are independent controls, when really they are both representing the same value; this could be unexpected for the user - if they are the same control then that implies that they should be in a single consistent location.
- Users do not have to navigate to a page that contains a branch selector to confirm which branch they are currently working on.
- Having a control in a common area might not be ideal in the immediate term (as we have very few things is version-controlled). But, ideally, we want to design a solution for the ultimate goal when more things are under version control. It would be good to choose a reasonable solution from EAP and all the way through, and not moving it around. Re-educated users about a new UI might cost more confusion down the track.
- Moving the branch selector to a central area in the early stage also minimises the changes we need to make later on the project.
- Let’s not move their cheese; design for where we want to be in the future rather than where we are now

OctopusServer/ConfigAsCode/BranchSelectorLocation.md
```diff
@@ -0,0 +1,41 @@
...
+
+- It's a single and consistent place for the user to learn where branch selecting/switching occurs = less to think about
+- It's less maintenance in our codebase (rather than scattering the Branch Selector throughout various components) = less to test and arguably less risk
+
+### Disadvantages
```
bonnietsang-octopus, line 38: **Issue:** Visual confusion
**Solution:** Show the branch selector in an enabled/ disabled state, depending on whether the current page is under version control. 
- If a page is version-controlled (process editor page, and soon the deployment settings page) visually enable the branch selector.
- If a page is not version-controlled, visually disable the branch selector. When users hover on the selector, have a tooltip to explain which area is currently version controlled, and highlight those pages on the sidebar nav.
- A visual design to come… (now baking)

bonnietsang-octopus (commented):
Hi Mark, I have added some thoughts on the advantages of the alternative approach and a solution to the disadvantage. Please let me know what you think. :)

bonnietsang-octopus (approved):
I have added advantages and disadvantages to approach #1. 
Also added a second approach.

gupta-kartik (approved):
LGTM

PR #38, opened 2021-02-15
Step Templates Documentation

YuKitsune:
Describes the changes to how Action Templates (a.k.a Step Templates) are serialised to the database and version control, and how we want to approach this problem in the future.


uglybugger (approved):
This looks like a good summary to me. Thank you for writing it up!

PR #46, opened 2021-05-25
Add document on persistence.

uglybugger:
This PR adds a write-up of our position on persistence within the Octopus Server codebase.

### DIFF
--- a/OctopusServer/Persistence.md
+++ b/OctopusServer/Persistence.md
@@ -54,7 +54,7 @@ using (projectScope.Push(spaceId, projectPath))
     runbook.DoTheThing();
 }
 ```
-You can compose a `ProjectPath` from either just a `ProjectId` or a `ProjectId` and `GitBranch`.
+You can compose a `ProjectPath` from either just a `ProjectId` (if a database-only project) or a `ProjectId` and `GitBranch` (if version-controlled).
 
 It is not necessary to resolve a document store directly from a container, begin a child `ILifetimeScope` or do any other trickery to change the project scope - just use the existing `IDocumentStore<TDocument>` from within the `using (...)` block and it will use the correct project scope.
 
--- a/OctopusServer/Persistence.md
+++ b/OctopusServer/Persistence.md
@@ -1,0 +1,94 @@
+# Persistence
+
+By and large, we adhere to the principle of [persistence ignorance](https://deviq.com/principles/persistence-ignorance).
+
+We intend for our persistence abstraction to completely decouple us from whichever underlying data storage mechanism actually records an entity. **This is not an abstract/purist position.** We actually have multiple persistence mechanisms (notably SQL and Git) and we switch dynamically between them. We also have different layers of caching, entity population, tracking and so on.
+
+**TL;DR: To access an entity, take a constructor-level dependency on an `IDocumentStore<TDocument>`** where `TDocument` is the type of entity you need to access.
+
+# Transactionality
+
+Our API consumers have a not-unreasonable expectation that API calls they make either succeed or fail atomically. To this end, we have a [unit of work](https://martinfowler.com/eaaCatalog/unitOfWork.html) implementation which allows us to coordinate operations across multiple different types of transaction.
+
+We generally have a single unit of work (UoW) per HTTP request. For background/long-running tasks, we have a UoW per logical operation.
+
+NOTE: Our intention is to revisit our long-running `TaskController` pattern to decrease the need for long-lived, in-memory state, but that's outside the scope of this document.
+
+Within a single unit of work, we might touch several different entity types from several different projects. Each of those projects will have
+its own persistence settings:
+
+- Some projects might be wholly in the database.
+- Some projects might be in their own Git repository.
+- Some projects might be in a Git monorepository shared with several other projects.
+
+All of this means that there is no one "transaction" to commit.
+
+We wrap a logical unit of work around this collection of (0-many) individual transactions. The unit of work is completed or abandoned automatically
+via aspects (filters, middleware etc.) and should not need to be addressed directly during the normal course of operations.
+
+# Custom queries
+
+The `IDocumentStore<TDocument>` interfaces exposes a `.Query()` method which returns an `IQueryable<TDocument>`. The expected approach to writing a custom query is to use a LINQ expression against that `IQueryable<T>`. We have implementations of `IQueryable<T>` for both SQL and Git.
+
+Where a custom query is reused in multiple locations, we use an extension method class on the `IDocumentStore<TDocument>` interface. For example, we have a [reusable mechanism for loading a `Project` via its slug](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Core/Features/Projects/ProjectDocumentStoreExtensionMethods.cs).
+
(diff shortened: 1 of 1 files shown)


OctopusServer/Persistence.md
```diff
@@ -0,0 +1,94 @@
...
+NOTE: Our intention is to revisit our long-running `TaskController` pattern to decrease the need for long-lived, in-memory state, but that's outside the scope of this document.
+
+Within a single unit of work, we might touch several different entity types from several different projects. Each of those projects will have
+its own persistence settings:
+
+- Some projects might be wholly in the database.
+- Some projects might be in their own Git repository.
+- Some projects might be in a Git monorepository shared with several other projects.
+
+All of this means that there is no one "transaction" to commit.
```
Waldo000000, line 24: Great point. Another FAQ we might expect: Is the UoW atomic? i.e. What happens if one transaction successfully commits but another (in the same UoW) fails?

OctopusServer/Persistence.md
```diff
@@ -0,0 +1,94 @@
...
+
+We wrap a logical unit of work around this collection of (0-many) individual transactions. The unit of work is completed or abandoned automatically
+via aspects (filters, middleware etc.) and should not need to be addressed directly during the normal course of operations.
+
+# Custom queries
+
+The `IDocumentStore<TDocument>` interfaces exposes a `.Query()` method which returns an `IQueryable<TDocument>`. The expected approach to writing a custom query is to use a LINQ expression against that `IQueryable<T>`. We have implementations of `IQueryable<T>` for both SQL and Git.
```
Waldo000000, line 31:
```suggestion
The `IDocumentStore<TDocument>` interface exposes a `.Query()` method which returns an `IQueryable<TDocument>`. The expected approach to writing a custom query is to use a LINQ expression against that `IQueryable<T>`. We have implementations of `IQueryable<T>` for both SQL and Git.
```

OctopusServer/Persistence.md
```diff
@@ -0,0 +1,94 @@
...
+
+Where a custom query is reused in multiple locations, we use an extension method class on the `IDocumentStore<TDocument>` interface. For example, we have a [reusable mechanism for loading a `Project` via its slug](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Core/Features/Projects/ProjectDocumentStoreExtensionMethods.cs).
```
Waldo000000, line 33:
```suggestion
Where a custom query is reused in multiple locations, we use an extension method class on the `IDocumentStore<TDocument>` interface. For example, we have a [reusable mechanism for loading a `Project` via its slug](https://github.com/OctopusDeploy/OctopusDeploy/blob/fb83083950a4bcac1d76ebcb6fc0b9250ba262b3/source/Octopus.Core/Features/Projects/ProjectDocumentStoreExtensionMethods.cs#L16).
```

OctopusServer/Persistence.md
```diff
@@ -0,0 +1,94 @@
...
+In these situations, you'll need to make it clear which space and project path you mean when you ask, for instance, for an `IDocumentStore<Runbook>` to fetch a runbook. To do that, take a dependency on an `IProjectScope` and structure your code like this:
+
+```
+using (projectScope.Push(spaceId, projectPath))
+{
+    var runbook = await runbookDocumentStore.GetAsync(runbookId, cancellationToken);
+    runbook.DoTheThing();
+}
+```
+You can compose a `ProjectPath` from either just a `ProjectId` or a `ProjectId` and `GitBranch`.
```
Waldo000000, line 57: Raises the question of: When can I/should I use "just a `ProjectId`", and which branch is assumed in that case?

OctopusServer/Persistence.md
```diff
@@ -0,0 +1,94 @@
+# Persistence
```
Waldo000000, line 1: 💡 Is there a natural place for this doco as a README.md right alongside the code, in https://github.com/OctopusDeploy/OctopusDeploy/tree/master/source/Octopus.Core/Persistence?

Waldo000000 (approved):
Nice, some really useful stuff there, easy to read too 👏 

Left some comments to consider; HTH! :shipit:

OctopusServer/Persistence.md
uglybugger: Good point. I'll add a thing.

OctopusServer/Persistence.md
uglybugger: If we could have a symlink, I'd have it in both :)

I think there's at least a place for a README.md in the code linking to the `Persistence.md` page in this repository.

OctopusServer/Persistence.md
uglybugger: Good point. Addressed in 2d34ce0e5298a4027eeea62ca2b97e3a80badf69.

OctopusServer/Persistence.md
uglybugger: Thing added in dd7df1464f253901473c18569b81f2fe17a5fabf.

OctopusServer/Persistence.md
uglybugger: (On my list before adding that is consolidating all the persistence-related files into a `Persistence` namespace - they're a bit spread out at the moment.

OctopusServer/Persistence.md
```diff
@@ -0,0 +1,108 @@
...
+
+# FAQ
+
+## I need to create/read/modify/delete a Foo entity. How should I do that?
+
+Take a constructor-level dependency on an `IDocumentStore<Foo>`.
+
+At present, if you modify something, you'll still need to call `await fooDocumentStore.UpdateAsync(foo, cancellationToken);` to tell the store that you've changed something. This requirement will go away (along with the `.Update(...)` and `.UpdateAsync(...)` methods). The safe assumption for now is that if those methods exist you should call them, and when they cease to exist everything will just continue to work.
+
+**IMPORTANT:** Please don't modify an entity if you don't want your changes persisted. At some point, persistence of any mutations will happen automatically, so even if you're not currently calling `.UpdateAsync(...)` the behaviour will soon change so that your changes will be picked up anyway.
```
zentron, line 69: I'd be interested to see a link out to this architectural discussion as moving away from explicit modifications has its own design discussions.

OctopusServer/Persistence.md
```diff
@@ -0,0 +1,108 @@
...
+
+You _are_ likely to encounter explicit knowledge of project scoping in the following scenarios:
+1. Iterating over collections of projects;
+1. During background task execution; and
+1. In message bus event handlers (which implement `IEventuallyHandle<TEvent>`).
+
+In these situations, you'll need to make it clear which space and project path you mean when you ask, for instance, for an `IDocumentStore<Runbook>` to fetch a runbook. To do that, take a dependency on an `IProjectScope` and structure your code like this:
+
+```
+using (projectScope.Push(spaceId, projectPath))
```
zentron, line 51: This section probably needs a bit of a DANGER callout that it should really be an exceptional case and the implementer should understand what the dependency chain is around the using block due to the nature of modifying our (lifetime-scoped)  project context "singleton".
I'm actually thinking we should enforce the adding of comments each time this tool is used.

zentron (approved):
looks good. is it worth adding a section on some of the decorators around the persistence classes to explain the what, why and how?

MarkSiedle (approved):
Well explained 👍

OctopusServer/Persistence.md
uglybugger: Added an advisory note in e0606c4a24d9b31aa386d76073dbe1bb45bc07c3.

PR #47, opened 2021-06-03
Added a proposal for optional/required types

TomPeters:
This PR presents a way we might model optional and required values in the inputs of our step packages.

View the rendered page: https://github.com/OctopusDeploy/Architecture/blob/step-ui/optional-values/Steps/Components/StepUI-OptionalValues.md

### DIFF
--- a/Steps/Components/StepUI-OptionalValues.md
+++ b/Steps/Components/StepUI-OptionalValues.md
@@ -28,28 +28,28 @@ type MyInputs = {
 There are a few useful bounded contexts to consider for this problem
 
 - **UI Inputs**: The inputs used by the Step UI
-- **API Inputs**: The inputs used by Octopus Server and passed to the API layer
+- **API Inputs**: The inputs used by the Octopus Server backend and passed to the API layer
 - **Execution Inputs**: The inputs used by the step executor
 
 # Scenarios
 ## Required values
 
-Values should be required by default.
-
-For example, for these inputs
+For these inputs
 ```ts
 type MyInputs = {
     foo: number;
 }
 ```
 
-the `foo` value should be required for all bounded contexts, since it should be invalid to pass an empty value like `undefined` at any of these points in time. The Schema Validation should assert that any step inputs provided through the API also conform to this requirement.
+the `foo` value should be required for all bounded contexts, since it should be invalid to pass an empty value like `undefined` at any of these points in time. The Schema Validation should assert that any values provided through the API also conform to this requirement.
 
 ## Optional values
 
 There are only two scenarios where optional values make sense
 - A value can be optional within all bounded contexts, including at execution time
 - A value can be optional in the UI inputs, but not in the API inputs, nor in the execution inputs.
+
+This section provides more details about these scenarios, and also talks about a third scenario that is not required.
 
 In these examples, consider a set of inputs like the following, where we want the `foo` property to be optional in some bounded contexts.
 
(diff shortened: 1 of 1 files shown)


andrewabest:
I think the approach presented makes the most sense - it is UI specific, and the UI owns the canonical representation of the input types, so it is OK to add additional things to it that we can strip away as it enters downstream bounded contexts.

One thing that concerns me is making it obvious that this is the approach we want authors to take. I could very easily see someone used to typescript running into the issue of not wanting something as a default value (realising they need to supply a default via `createInitialInputs`), and their natural reaction would be to make it optional, or undefined. They would probably then just deal with narrowing this back down at execution time. 

There wouldn't be anything wrong with doing that - and I suspect a lot of people would not realising there is a better approach.

At least in this case authors can get to a solution either way - one way is just a bit nicer, as long as they know about it.

TomPeters:
Thanks @andrewabest.

Do you have any concrete thoughts/preferences regarding using a type like `EmptyInitialValue` compared to something more general like `UIInput<T>`?

> One thing that concerns me is making it obvious that this is the approach we want authors to take. I could very easily see someone used to typescript running into the issue of not wanting something as a default value (realising they need to supply a default via createInitialInputs), and their natural reaction would be to make it optional, or undefined. They would probably then just deal with narrowing this back down at execution time.

>There wouldn't be anything wrong with doing that - and I suspect a lot of people would not realising there is a better approach.

Agreed - I don't see this as being a huge issue as things would still work, but the code would be a little less nice. I can think of two things that might help guide them towards this proposed option
- They would probably want to write a validation rule to assert that the input value is not undefined, since it must be selected. They would then also be forced to check (in order to type narrow) the value in the step executor. Having to do this in two places might trigger some people's spidey senses and make them look for a different way of doing it.
- I'm trying to imagine what the API would look like for a Select/RadioButtons control that allowed having no value selected by default. Perhaps it is something like this:

```
radioButtons({
    input: inputs.myRadioButtonsInput,
    options: [...],
    allowNoInitialSelection: true,
})
``` 

Perhaps if `allowNoInitialSelection` is true, the type constraints force the `EmptyInitialValue` type to be assignable to `inputs.myRadioButtonsInput`. They would receive a type error indicating that they didn'n include `EmptyInitialValue` in their set of possible types. If they remove `allowNoInitialSelection`, they will find (at runtime? or perhaps as part of our build cli tool?) that some empty initial state they have provided doesn't actually allow the radio buttons to show that nothing is selected.

TomPeters:
>One thing that concerns me is making it obvious that this is the approach we want authors to take. I could very easily see someone used to typescript running into the issue of not wanting something as a default value (realising they need to supply a default via createInitialInputs), and their natural reaction would be to make it optional, or undefined. They would probably then just deal with narrowing this back down at execution time.

>There wouldn't be anything wrong with doing that - and I suspect a lot of people would not realising there is a better approach.

The other thing I like about this is that we can roll with this solution for now, and add `EmptyInitialValue` at a future time.

andrewabest:
> Do you have any concrete thoughts/preferences regarding using a type like EmptyInitialValue compared to something more general like UIInput<T>?

I feel like having a UI-only type is confusing - we don't really want people meta-programming with these abstractions, just programming :) I like the semantics of `EmptyInitialValue`, and I think sticking to solving the problem we know and not generalising to solve problems we don't know about is a good idea for now.

> The other thing I like about this is that we can roll with this solution for now (i.e. needing to check the value at execution time and write a validation rule), and add EmptyInitialValue at a future time when we find it valuable.

TomPeters:
>Big +1 here, having stated it I feel like this optimisation is not high priority. It is good to know the way we want to go, but for now I'd be happy to live without it.

Agreed. I'm glad that we ruled out more advanced or complex things like "parsing instead of validating". I think the only missing piece in the short term might be adding support for union types in our inputs schema, so that we can actually write types like `number | undefined` or `myProp?: number`.

We only currently support unions of object, not primitive types ([source](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/newportal/app/components/StepPackageEditor/StepPackage/Schema/convertFromJsonSchemaToInputSchema.ts#L78)), and we also don't really support null or undefined (I _think_ these would be represented by `required` properties in the json schema, [which is explicitly removed right now](https://github.com/OctopusDeploy/step-api/blob/main/packages/cli/src/commands/SchemaGenerator.ts#L60-L62)).

PR #48, opened 2021-06-06
New steps guidelines

mcasperson:


Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,250 @@
...
+    * Are microservices for a related service likely to be deployed in many regions or availability zones? This is likely for high availability. But is this better modeled as a single step with multiple targets, or a duplicated step with overridden AZ fields? Clearly duplicating steps to override one field will result in a mess, and multiple targets are a better choice here.
+    * Are teams likely to deploy a feature branch to a new GCP project? This is likely given project scoped resources like GAE routing rules.
+* The names of services, such as a Lambda function name, an Azure Container Instance name, or a Google Cloud Run instance, will typically be overridable. Feature branch deployments will almost certainly require renaming these values on a step.
+* There is no way to get the union of two targets. For example, the deployment of a Lambda exposed by an API Gateway instance can not combine the details of a Lambda target and an API Gateway target. Where a deployment takes place to two services that could be targets in their own right, consider how a single target can lift those combined details from a step.
```
TomPeters, line 67: This feels like pragmatic advice given where we are today :+1:, but it makes me think about what the next steps are for Octopus.

There's been talk in the past about giving Octopus good visibility into your infrastructure, imagining what it would be like if Octopus could generate a map of all of your infrastructure.

As soon as you start overriding fields in deployment steps, you lose a lot of the required information for that kind of capability.

This line of thinking makes me ask the question: What changes could we make to the way we model things so that feature branching and microservice deployment scenarios could be reasonably modelled with everything as a deployment target?

Yes this might still involve an explosion of targets, but is that really an issue? Could we make that a non-issue?

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,250 @@
...
+
+To identify old resources, anything that can be tagged or labeled should include the following keys (example values are also shown) by default:
+
+* `Octopus.Project.Id`: "projects-1"
+* `Octopus.Action.Id`: "8427fd24-17c8-4c7d-b32f-d2b8d51f2121"
+* `Octopus.Deployment.Id`: "8427fd24-17c8-4c7d-b32f-d2b8d51f2121"
+* `Octopus.RunbookRun.Id`: "runbookruns-86"
+* `Octopus.Step.Id`: "b6f8fd75-3acf-4186-a573-285a1aa11a9a"
+* `Octopus.Environment.Id`: "environments-4"
+* `Octopus.Deployment.Tenant.Id`: "untenanted"
```
TomPeters, line 174: Should we also tag with human readable names too?

E.g. additionally adding `Octopus.Project.Name`.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,250 @@
...
+* Proposed Scope - This is the opposite of the constraints, and provides an overview of the features to be delivered. This section should link to the blog post and the UI table document.
+* Future Considerations - Most pitches expect to be followed up with another related pitch with the next milestone of work. This section provides a very high level list of features that might be considered in the next milestone.
+
+## UI table
+
+This document details the proposed UI structure of the new steps. This is useful when collaborating with the design team to find any unique UX patterns that may need to be incorporated into Octopus:
+
+![](assets/uimockup.png)
+
+An example of this document can be found [here](https://docs.google.com/document/d/13jZVn_L6U3_iFEy8W12NO2TochPEu_g_ZNEjo8R652o/edit).
```
TomPeters, line 229: I'm still concerned that this approach is not collaborative enough, and tries to front-load too much of the design work.

For example, the development team might end up taking things in a different direction, meaning that the design work established in this document was wasted or incorrect.

Could this instead just be a part of working on the pitch during the build phase of a cycle? If we start with small pitches and small incremental milestones, I see a decreasing need for a large amount of upfront design work.

Steps/PitchGuidelines.md
mcasperson: That's a good idea. I'll add the names of relevant things to the list.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,267 @@
...
+## Remember to update associated projects
+
+New steps and targets will need to be reflected in other projects:
+
+* Octopus client - this may need to be updated with new target types. This requirement may be removed with step packages.
+* Dynamic target scripts - the [docs](https://octopus.com/docs/infrastructure/deployment-targets/dynamic-infrastructure) will need to be updated with any new scripts.
+
+# Pitch documents
+
+The pitch process generates a number of documents, each with it's own use case and audience. These are detailed below.
```
rhysparry, line 214: its

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,267 @@
...
+
+## RFC blog post
+
+The first document is a Request For Comment (RFC) style blog post. The purpose of this post is to:
+
+* Inform customers, partners, and other internal Octopus departments of the proposed new functionality.
+* Highlight the details of the new functionality, calling out how it expresses our vision of best practice deployments.
+* Include enough UI mockups to provide a sense of what the feature looks like.
+* Provide a link to a [GitHub issue](https://github.com/OctopusDeploy/StepsFeedback/issues) where people can provide feedback.
+* But never promise to deliver any specific feature. Always call out that the feature has not been committed to, and that there is no release date.
```
rhysparry, line 224: Should we call out the proposed scope hammering for initial milestones? What about subsequent milestones?

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,267 @@
...
+
+The pitch document is based on the [Shape Up](https://basecamp.com/shapeup/webbook) process.
+
+This document is structured with the following sections:
+
+* Executive Summary - two or three paragraphs outlining how this pitch aligns to the company's strategic goals, why this particular functionality is important, and a high level overview of the proposed solution.
+* Problem - A description of the problems faced by customers that this pitch will solve.
+* Constraints - Each pitch relates to a milestone, and milestones have been constrained to deliver the smallest unit of useful functionality. This section lists the constraints that exclude work to be done.
+* Proposed Scope - This is the opposite of the constraints, and provides an overview of the features to be delivered. This section should link to the blog post and the UI table document.
+* Future Considerations - Most pitches expect to be followed up with another related pitch with the next milestone of work. This section provides a very high level list of features that might be considered in the next milestone.
```
rhysparry, line 236: Something I found when writing the pitch was that the "Constraints" and "Future Considerations" sections are very closely related (future considerations tend to be about loosening constraints).

I think this is generally fine, especially as it starts to form pieces of work for future milestones. However, maybe it makes it harder to identify constraints that are not planned to change later?

Steps/PitchGuidelines.md
rhysparry: I think the separation of the documents will actually help here. Perhaps this is a document that is more typically a skeleton at the beginning of the development process?

It was definitely helpful to have the document explicitly call out what fields are part of the planned scope.

This document seems to be the one that is most open to change all throughout the process. It's more of a communication tool between Product, Design and Development than anything else.

rhysparry (commented):
Some of these are tangents

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,250 @@
...
+
+We are leaders in the deployment space. We have years of experience and our tools have performed millions of deployments, so customers look to us to provide clear opinions on what best practice deployments look like. These values have been distilled into [The ten pillars of pragmatic deployments](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments).
+
+The steps we deliver must have a clear vision for how they enable teams to deliver best practice deployments. Often this involves combining, hiding, not supporting, or hard coding some features in the platforms we deploy to.
+
+Where there is no clear decision on an opinion, refer to **Ship incrementally** for guidance.
+
+## Graduated path from opinionated to raw scripts
+
+Highly opinionated steps will fail some of our customers all of the time. To support those with advanced use cases, steps should provide a graduated path from opinionated (and often monolithic) steps, to granular and composable steps, to raw templates or scripts.
```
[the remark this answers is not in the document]
mcasperson: 👍 Splitting out the composable steps (i.e. task definition and service) as well as conversion directly to a template would be a great feature. It is something we need to consider for future ECS milestones.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,250 @@
...
+* AWS App Runner
+* GCP App Engine
+* GCP Cloud Run
+* Azure container instances
+
+![](assets/secondlayerservice.png)
+
+### Third layer service
+
+A third layer service has a many to one relationship with the cloud provider, and a one to many relationship with the deployments it holds. Examples include:
```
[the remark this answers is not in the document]
mcasperson: I'm not sure I've ever seen something like that. Do you have an example service you were thinking of?

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,267 @@
...
+
+These tags have proven useful for identifying resources related to previous deployments, regardless of environments, tenants etc.
+
+## Output variables
+
+Many deployments will create resources that need to be consumed by subsequent steps. The following are guidelines for output variables:
+
+* Links to any resources accessible via a URL. This helps with testing, and speaks to the [verifiable deployments](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments#verifiable-deployments) pillar.
+* Names of any generated resource identifiers.
+* Details of any generated revision numbers.
```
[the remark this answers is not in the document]
mcasperson: 👍 Yes, I'd expect these to be defined as output variables from whatever template language we build the resources from.

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
rhysparry: Azure functions perhaps? So the "Function App" is the target and there is a limit of 1 additional slot (with the consumption plan). You might have multiple functions under the function app, but the individual deployment unit is the entire function app.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,267 @@
...
+
+Where a target platform has functionality that prevents advanced deployment patterns, we'd likely offer an opinionated step that does not use the conflicting functionality so as to promote deployments that enable advanced patterns. An example of this is [AWS API Gateway](https://octopus.com/blog/deploying-lambdas#why-limit-ourselves-to-one-stage-per-environment), whose stages prevent hotfixes and feature branches.
+
+### Calculating networking rules
+
+Watch for platforms that require knowing the current state of network rules configured with previous deployments. For example, many network rules use relative weights for directing traffic e.g. service1 has a traffic weight of 10, service2 has a traffic weight of 20. If we deploy a service3, any weighted value is entirely relative to the weights assigned to the other services, and the outcome is not repeatable.
+
+We should aim to express network traffic as percentages. This may mean translating a percentage into the appropriate weight at deployment time. A psudeocode algorithm for this is:
```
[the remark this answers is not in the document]
mcasperson: Nice catch, I'll fix this.

Steps/PitchGuidelines.md
mcasperson: 👍

Steps/PitchGuidelines.md
mcasperson: That is a good idea. I'll add a point to call out the constraints and whether they will be addressed in future milestones.

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
mcasperson: Patching templates generated by the UI is an interesting idea. We tied to make something like Kustomize work with Kubernetes templates, but it just never quite fit anywhere. This was actually one of the reasons why we implemented the idea of converting opinionated steps to templates.

I'm hopeful that faster release cycles decoupled from Octopus core mean we could address requests to add new fields in days rather than weeks or months.

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
mcasperson: I see - I was calling these second layer services, as in Azure itself (first layer) orchestrates Function Apps (second layer) that we then deploy into.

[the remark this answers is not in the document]
rhysparry: I thought second layer services had the restriction that there could only be one in a particular region/project. Happy to treat them that way. I guess the question I have now is, how this categorisation affects our approach.

The key differentiator that I'm seeing is that some targets can host many "things", while some targets only host one "thing".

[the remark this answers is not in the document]
mcasperson: This is a tricky area because the differences are subtle, but affect how we model steps and targets.

The "thing" I was trying to capture was the overarching orchestrator service of a target. I'll try and update the text to make that clearer.

For example, there is only one global Lambda orchestrator per AWS region, and only one Google App Engine orchestrator per project. Likewise (excluding resource groups) there is only one orchestrator for web or function apps in Azure.

The main difference is in how steps and targets work.

For third layer services, the step itself tends to create the resources that run the deployment, and so those resources are not defined by the target. For example, the Kubernetes target doesn't have a default name for a deployment resource, and ECS target doesn't have a default name for a service. These services already avoid target explosion by virtue of the fact that resources are created in the steps.

For second layer services, we do find resources named in targets. For example, an Azure web app target names the Web App instance. The deployment steps also do not expose the ability to create a web app instance as part of the deployment.

Targets for other second layer services like Lambda or App Runner would be represented like Azure web app targets. They would include the credentials, region, project, and default name of the Lambda or App Runner service.

The question really comes down to what values make sense to put in the target?

It never made sense to try and put default resource names on the Kubernetes target, because a single deployment could involve any number of configmaps, secrets, volumes etc. And with Custom Resource Definitions (CRDs) we simply couldn't know the resources that we were providing default values for.

I think you could argue either way for ECS. ECS only supports a few different types of deployment resources like services, tasks and scheduled tasks, so maybe default values for those could be put on a target as overridable fields. I'm comfortable treating an ECS target like a K8s target and leaving it to the steps to name and create these resources though, but it might be something to discuss.

The real question is what values would appear on an AWS Lambda or App Runner target? Both of these are second layer services, and so would only differ by the default name of the Lambda or App Runner service. 

If we left this name off the target, we'd just have a generic AWS target. This is not ideal though as looking forward we want targets to offer richer service specific functionality, and ongoing discussions around concepts like dynamic targets would benefit from a rich selection of first-class Octopus target types.

If we put a default value on a target, we lift the “where” from the step, but need to make the values overridable so things like feature branches don’t result in target count explosion.

Although maybe instead of second and third layer services, a better distinction are services that create the resources executing the deployed code dynamically (ECS, K8s, Lambda) and those that don’t (web/function apps)?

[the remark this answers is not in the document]
mcasperson: Actually I think maybe second and third level are the wrong abstractions. I'll take another pass at this to try and capture the target differences another way.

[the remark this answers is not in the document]
mcasperson: I've provided a different way of explaining why some targets have fields overridden by the steps and some don't. I think with this new classification system, we would probably include a default service name on an ECS target. Does this new classification make more sense @rhysparry ?

Steps/PitchGuidelines.md
mcasperson: We will need some design work to be done up front for the RFC blog post. I imagine the more complex elements of a new step would be what we want to show off in the post to get people interested in the feature and to provide feedback.

It is also valuable to capture complex UI elements with no existing patterns as part of exploring the kind of rabbit holes we'll have to call out in the pitch.

However I would expect in future that these documents would be created by the developers and designers together, which should mitigate the problem of both teams going in different directions. If a team lead was to create this document, would that alliterative some of your concerns @TomPeters?

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Combine loosely coupled externalized configuration with the deployment
+
+Deployments will often rely on externalized configuration. 
+
+In some cases the configuration is tightly bound to the deployment. For example, environment variables are often defined as part of a deployment, and tightly bound to the application they are configured against.
+
+In other cases two loosely coupled resources will combine to define a deployment. For example, a Kubernetes Pod may reference a ConfigMap for environment variables.
+
+Where two loosely coupled resources must exist side by side for a deployment to operate in a predictable manner, consider an opinionated step that deploys both. This will help customers fall into the pit of success regardless of whether they are using advanced deployment patterns like canary or rolling deployments, or performing a rollback.
```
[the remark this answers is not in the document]
mcasperson: This is a good point. Combing resources assumes that they are _instantiable_, which has now been defined under the _Use Targets_ section.

I don't think Azure resources fit the definition of _instantiable_, because you can not simply provide a blob or queue name and have one created since they need an associated storage account.

Feature branch deployments of Azure functions would have to consider adding dedicated steps to create blobs and queues, or document how these resources would be created with ARM or Terraform templates.

Steps/PitchGuidelines.md
TomPeters: >We will need some design work to be done up front for the RFC blog post.

:+1: While true, I would expect the design here to be minimal, even a wireframe is probably fine. I think the underlying principle is the same as using a "fat marker" sketch in pitches: It is more about getting the general idea across, and less about illustrating a specific design.

>It is also valuable to capture complex UI elements with no existing patterns as part of exploring the kind of rabbit holes we'll have to call out in the pitch.

True, but the cost of not capturing these is probably low. The difference between adding a new complex widget and re-using the ones we already have is UX. It is viable to ship something simple for an early milestone using the components we have available, and then improve the UX (e.g. add a new type of component) as part of a future milestone. This means we get to discover the requirement together as we are working on implementing the step, which feels like a more appropriate time.

>However I would expect in future that these documents would be created by the developers and designers together

:+1:

>If a team lead was to create this document, would that alliterative some of your concerns @TomPeters?

Yes that helps a bit, but _only_ if it is created at build time.

I think my ideal workflow would look something like this:
- We identify that we need to make a change (e.g. we need to add support for network configuration for azure containers).
- We work on getting a walking skeleton set up: Add a basic inputs (like port); Use this in the step executor to configure the network.
- Either at the same time as setting up the walking skeleton or immediately afterwards, we decide what options we want to support and how we need to do it. This would involve research, collaboration with product, collaboration with design, etc.
- We enhance the walking skeleton with those changes.

The advantages of this approach are
- We have a walking skeleton implemented which helps us find unknowns. This means our decision making happens at the point in time when we have as much information as possible.
- We can test and iterate on our ideas very quickly. Adding a new field and seeing how it looks in the UI should take seconds. Using new fields in the step executor and seeing if it works shouldn't take much longer.
- By this point in the process (during the build phase), everyone (product/design/engineering) should have enough context to contribute meaningfully.
- We have a working UI that we can collaborate on, rather than needing a separate document. The code is the source of truth. The UI framework should also be simple enough that a designer or someone from product can contribute to it.

All that being said, we haven't really implemented a new step using the step package infrastructure from scratch yet, so I actually think it might be better to not be prescriptive at this time, and instead we work out a new process as we go.

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
rhysparry: I'm not sure whether default service names on ECS make sense, mostly because of its similarities to Kubernetes. I'm open to it, but it doesn't "feel" right.

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
rhysparry: I think we need to better understand how people will actually use functions and their expectations around deployment. Do the patterns for large deployments fit if you are only deploying one component in a larger system? Are there better patterns?

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
mcasperson: 👍 I think we could fall back on the guidelines for shipping incrementally and lean towards not adding a field initially.

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
mcasperson: I agree, we're somewhat blind to actual use cases here. I'm hopeful that we can specifically ask for that kind of feedback in the RFC blog post and get some actionable comments.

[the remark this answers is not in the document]
rhysparry: There may even be a case for an initial reconnaissance RFC post. In such a post we wouldn't necessarily be postulating a solution, just discussing the problem space and maybe some initial thoughts. This is somewhat dependent on the quality of this as a feedback channel 🤞.

andrewabest:
I'm intending to review this document in sharpening time (next week)

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+Where there is no clear decision on an opinion, refer to **Ship incrementally** for guidance.
+
+## Graduated path from opinionated to raw scripts
+
+Highly opinionated steps will fail some of our customers all of the time. To support those with advanced use cases, steps should provide a graduated path from opinionated (and often monolithic) steps, to granular and composable steps, to raw templates or scripts.
+
+This path allows us to express our opinions regarding best practice deployments with opinionated steps that will often merge many underlying platforms and resources. The granular steps provide the ability to compose deployments in unique ways, while still retaining the benefits of a UI driven approach. Raw templates or scripts provide the ultimate level of customization, free of the opinions baked into the specialized steps.
+
+## Use targets
```
andrewabest, line 34: I feel like the level of detail we dive into for Targets here obscures the purpose of the document, which is more to call out what people should consider when designing and refining new steps, and what should be considered pre/during/post delivery as well. Could we pull out the detail for deployment targets into a new document (pretty much every L3/4 heading below `Use Targets`, and link to that here? Perhaps we could call it `TargetGuidelines` or something similar.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
+# Introduction
+
+Neutrality is one of the selling points of Octopus. We aim to support all the most popular cloud native platforms, while retaining a consistent user experience and common best practices. To achieve this, all new steps (which will typically be cloud steps) in Octopus should aim to implement a consistent set of base functionality. 
+
+This document provides guidelines for pitching steps that are consistent, implement the lessons we have learned over the years, and express our opinions on best practice deployments. It also captures some processes to be followed once a pitch has been delivered.
```
andrewabest, line 5:
```suggestion
This document: 
- Provides guidelines for pitching steps that are consistent and express our opinions on best practice deployments
- Captures some processes to be followed once a pitch has been delivered.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
+# Introduction
+
+Neutrality is one of the selling points of Octopus. We aim to support all the most popular cloud native platforms, while retaining a consistent user experience and common best practices. To achieve this, all new steps (which will typically be cloud steps) in Octopus should aim to implement a consistent set of base functionality.
```
andrewabest, line 3:
```suggestion
Neutrality is one of the selling points of Octopus. We aim to support all the most popular cloud native platforms, while providing a consistent user experience and common best practices. To achieve this, all new steps in Octopus should aim to implement a consistent set of base functionality. 
```

andrewabest: We do mention our aim above, so just trying to simplify some of the messaging in the doc

Steps/PitchGuidelines.md
andrewabest: Simplifying

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+This document provides guidelines for pitching steps that are consistent, implement the lessons we have learned over the years, and express our opinions on best practice deployments. It also captures some processes to be followed once a pitch has been delivered.
+
+# Pitch guidelines
+
+The following guidelines capture the functionality and opinions that we should apply to all steps.
+
+## Ship incrementally
+
+Team steps has devoted a great deal of effort to building a framework allowing features to be shipped independently of the main Octopus code base. This allows us to ship features more frequently, and respond to user feedback more quickly.
```
andrewabest, line 13:
```suggestion
The Steps team has devoted a great deal of effort to building the Step Package architecture, which allows features to be shipped independently of the main Octopus Server code base. This allows us to ship features more frequently, and respond to user feedback more quickly.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+To take advantage of this, pitches should be broken down into milestones. Each milestone should:
```
andrewabest, line 15:
```suggestion
To take advantage of this, pitches should be broken down into **milestones**. Each milestone should:
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+* Deliver enough value to be useful for our customers.
```
andrewabest, line 17:
```suggestion
* Deliver enough value to be useful for our customers.
* Be shippable in 4 weeks or less.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Be opinionated
+
+We are leaders in the deployment space. We have years of experience and our tools have performed millions of deployments, so customers look to us to provide clear opinions on what best practice deployments look like. These values have been distilled into [The ten pillars of pragmatic deployments](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments).
+
+The steps we deliver must have a clear vision for how they enable teams to deliver best practice deployments. Often this involves combining, hiding, not supporting, or hard coding some features in the platforms we deploy to.
+
+Where there is no clear decision on an opinion, refer to **Ship incrementally** for guidance.
+
+## Graduated path from opinionated to raw scripts
```
andrewabest, line 28:
```suggestion
## Graduated path from opinionated steps to raw scripts
```

Steps/PitchGuidelines.md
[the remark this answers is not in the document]
andrewabest:
```suggestion
Highly opinionated steps will not suit all of our customers. To support those with advanced use cases, steps should provide a graduated path from opinionated steps, to granular and composable steps, to raw templates or scripts.
```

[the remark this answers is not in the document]
andrewabest: Simplifying

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Use targets
+
+Targets must be used to define where a deployment takes place. They help ensure deployment steps are decoupled from their final destination, which speaks to the [repeatable deployment](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments#repeatable-deployments) pillar.
+
+Targets should capture:
+
+* The credentials required to perform a deployment.
+* The default worker a deployment will be performed on.
+* Cloud specific details like regions, projects, and resource groups.
```
andrewabest, line 42:
```suggestion
* Cloud specific details that locate the target like regions, projects, and resource groups.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+* The name of the service being deployed to e.g. the name of a ECS cluster, an Azure Web App, a Lambda function etc.
```
andrewabest, line 43:
```suggestion
* The identifier of the service being deployed to e.g. the name of a ECS cluster, an Azure Web App, a Lambda function etc.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+* The name of any nested partitioning e.g. namespaces, slots etc.
```
andrewabest, line 44:
```suggestion
* Identifiers of any nested partitioning e.g. namespaces, slots etc.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Overrideable and non-overridable fields
+
+We aim to strike a balance between the flexibility of allowing target fields to be overridden on steps, and the pragmatism of encouraging target use as the foundation of repeatable deployments. This is often subjective, so there is no hard and fast rule here.
+
+Targets provide the ability to lift deployment destination details out of the steps, and represent a security boundary within Octopus. For those teams that require a high level of control or a clear distinction between deployment processes and the infrastructure they are deployed to, the solution is to lift all details about the deployment destination into a target.
+
```
andrewabest, line 51: I agree with Tom that it might be nice to add a paragraph in here talking about our intended future state with deployment targets - giving customers visibility over their entire deployment topologies.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+However, without the ability to override target details on a step, common deployment scenarios like feature branching and microservice deployments will result in an explosion of targets, where the individual targets provide little benefit. In these situations it is useful to consider which target fields must be overridable in steps to support these deployment patterns.
```
andrewabest, line 52: Looking at [The pipelines pitch here](https://docs.google.com/document/d/184_iYwPtctj5X3a44DdFC-T71dc2nO0vHp1GEtdpbxw), we can see an idea of having a `features` pipeline, and a `default` pipeline. 

Deployment processes are defined independently to pipelines. In the example, feature branches use the same process as deployments up to production. 

Pipelines are where Environments are associated.

The pitch talks about improving `Step Target Roles`, suggesting that since we know which environments are associated with which pipelines, and which pipelines are referencing our process, we can derive which set of targets may be selectable within the role selection for each step.

The interesting thing here is that we get a closer look at which targets will be associated with a step. Steps should know what targets that they run on - there is always a 1:1 affinity (with a couple of odd exceptions) between a step and a target type.

> In these situations it is useful to consider which target fields must be overridable in steps to support these deployment patterns.

What if we lifted this up to first-class functionality, where any step could indicate that it wanted to override one or more fields for its target? If we had that, we wouldn't need to decide which fields should be overridable - you could just override any of them (probably with the exception of accounts)

It feels like this sort of overridable functionality makes the most sense for dynamic infrastructure, that we want to merge branch names into target identifiers for entropy to avoid clashes when creating resources. 

I agree with @TomPeters that it feels like in general we don't want to encourage people to be overriding deployment target values just to keep the infrastructure page smaller - that feels like a UX challenge to tackle instead. 

Looking a bit further ahead - it feels like for things like feature branch deployments, we want to create sets of deployment targets, and then to be able to view them as a set - i.e. here is the topology currently deployed for this particular release of project X. This view could be given within the project boundary, instead of outside of it. We could provide this view when you looked at the details of a particular release. Perhaps there would be a way to segment off this infrastructure from the "global" infrastructure view, if that is a desired property.

There is a bit of thinking here, happy to progress this outside of this PR.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+Here are some guidelines:
+
+* When in doubt, assume targets describe the **where** and the **who**. Steps define the **what** and the **how**.
```
andrewabest, line 56: I'm not sure if this guideline really helps - I'm not sure how to action it. Can we make it a bit clearer? I feel like you'd need to share a mental model for this to make sense. Perhaps an example may help as a sub-point below?

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+    * Are microservices for a related service likely to be deployed in many regions or availability zones? This is likely for high availability. But is this better modeled as a single step with multiple targets, or a duplicated step with overridden AZ fields? Clearly duplicating steps to override one field will result in a mess, and multiple targets are a better choice here.
+    * Are teams likely to deploy a feature branch to a new GCP project? This is likely given project scoped resources like GAE routing rules.
+* The names of services, such as a Lambda function name, an Azure Container Instance name, or a Google Cloud Run instance, will typically be overridable. Feature branch deployments will almost certainly require renaming these values on a step.
+* There is no way to get the union of two targets. For example, the deployment of a Lambda exposed by an API Gateway instance can not combine the details of a Lambda target and an API Gateway target. Where a deployment takes place to two services that could be targets in their own right, consider how a single target can lift those combined details from a step.
+
+### Types of target fields
```
andrewabest, line 69:
```suggestion
### Target identification
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+There are two ways to classify fields that we might consider placing on targets.
```
andrewabest, line 71:
```suggestion
The fields on targets exist for us to identify a specific target we want to act upon during a deployment. There are two ways to classify fields that we might consider placing on targets.
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### Instantiable
+
+The second classification is whether the resources are **instantiable**. A resource is instantiable if the act of deploying an application usually also creates the resource that hosts it. Or, in other words, instantiable resources are self contained.
+
+Many resources are instantiable. For example, providing a name for a Kubernetes resource, AWS Lambda, AWS App Runner service, Google App Engine service, or Google Cloud Runner service, along with the values usually supplied while deploying the application, is all that is required for that resource to be created during a deployment if it does not exist.
+
+Services like Azure web or function apps do not create the underlying resources with a name. An app service needs a resource group, networking, service plan etc. So simply supplying a name, along with the details usually associated with an application deployment, can not create a web app instance if it doesn't exist.
+
+#### Target values
```
andrewabest, line 91:
```suggestion
#### Target fields
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+The following table provides guidelines for including resource names on targets:
+
+| Knowable    | Instantiable  | Target field type | Health Check |
+|-------------|---------------|-------------------|--------------|
+| Yes  | Yes  | Optional value that can select existing resource. | Verify credentials, don't assume resource exists. |
```
andrewabest, line 97:
```suggestion
| Yes  | Yes  | Optional field that can select existing resource. | Verify credentials, don't assume resource exists. |
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+| Yes  | No   | Mandatory value that must select existing resource. | Verify credentials and verify resource exists. |
```
andrewabest, line 98:
```suggestion
| Yes  | No   | Mandatory field that must select existing resource. | Verify credentials and verify resource exists. |
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### Overridable and uninstantiable target values
+
+Where a target field references a resource that is uninstantiable, it usually means it is a non-trivial task to create a new instance.
+
+Where a target field can be overridden by a step, it usually means that deployment patterns like feature branches benefit from deploying to a new resource.
+
+Where a field references a uninstantiable resource that can be overidden on a step, consider how such resources can be created as part of a deployment. It may be a case of documenting a Terraform, CloudFormation, or ARM Template that creates the necessary resources. Or it may be worth providing custom steps that create these resources.
+
+## Watch for changing variables
```
andrewabest, line 109:
```suggestion
## Immutable resources and re-use
```

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+All targets should support the ability to inherit credentials from the worker a deployment is executed from, and avoid the need for long lived credentials to be maintained by Octopus.
+
+Examples include [AWS EC2 IAM roles](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html), [Azure Managed Service Identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview), [Google inherited service accounts](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances), and [Kubernetes pod service accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/).
+
+## Declarative over imperative
+
+All cloud providers and modern orchestration platforms offer a template language for defining resources. They also provide a great deal of tooling and reporting for these "managed" resources. This speaks directly to the [auditable deployments](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments#auditable-deployments) pillar.
+
+Steps should prioritize the creation of resources through declarative templates rather than imperative CLI or SDK commands.
+
```
andrewabest, line 132: Should we add a paragraph here about how this also supports "Graduated path from opinionated to raw scripts" above?

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+Many cloud native platforms have built in support for blue/green or canary deployments. If so, we want to expose these as part of an Octopus deployment where it makes sense. This speaks to the [seamless deployments](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments#seamless-deployments) pillar.
+
+### Prioritize seamless deployments over platform limitations 
+
+We also want to provide support for other deployment scenarios like hotfixes and feature branches. These are often not natively supported by the target platforms.
+
+Where a target platform has functionality that prevents advanced deployment patterns, we'd likely offer an opinionated step that does not use the conflicting functionality so as to promote deployments that enable advanced patterns. An example of this is [AWS API Gateway](https://octopus.com/blog/deploying-lambdas#why-limit-ourselves-to-one-stage-per-environment), whose stages prevent hotfixes and feature branches.
+
+### Calculating networking rules
```
andrewabest, line 143: This feels like very specific guidance, to the point of implementation detail. I'd suggest we can remove this particular chunk - it is something our teams should be able to identify and pick up when developing any steps that require network traffic weighting.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+These tags have proven useful for identifying resources related to previous deployments, regardless of environments, tenants etc.
+
+## Output variables
+
+Many deployments will create resources that need to be consumed by subsequent steps. The following are guidelines for output variables:
+
+* Links to any resources accessible via a URL. This helps with testing, and speaks to the [verifiable deployments](https://octopus.com/blog/ten-pillars-of-pragmatic-deployments#verifiable-deployments) pillar.
+* Names of any generated resource identifiers.
+* Details of any generated revision numbers.
+
```
andrewabest, line 219: Should we group `Documentation` and `Updating projects` under another L1 heading of # Post-Delivery or something? It feels like segmenting them off would be helpful.

Steps/PitchGuidelines.md
```diff
@@ -0,0 +1,287 @@
...
+All new features must be documented. Ensure this is a required feature of any pitch being developed.
+
+## Remember to update associated projects
+
+New steps and targets will need to be reflected in other projects:
+
+* Octopus client - this may need to be updated with new target types. This requirement may be removed with step packages.
+* Dynamic target scripts - the [docs](https://octopus.com/docs/infrastructure/deployment-targets/dynamic-infrastructure) will need to be updated with any new scripts.
+
+# Pitch documents
```
andrewabest, line 231: It feels like we have a couple of documents in one here - one is about designing steps (and really has a bunch more detail that we would add to the pitch document itself), and one is about delivering steps - the process and artifacts, and post-delivery things that need to be done. 

What if we split this document into three - one for `Steps/DeliveryGuidelines.md`, one for `Steps/StepDesignGuidelines.md`, and one for `Steps/TargetDesignGuidelines.md`? It may make each of them easier to digest, and allow us to make each a bit more focussed.

Steps/PitchGuidelines.md
mcasperson: That is a great point. I'll add some specific examples of these to help make it clearer.

Steps/PitchGuidelines.md
mcasperson: That makes sense. I'll call out this relationship between the guidelines.

Steps/PitchGuidelines.md
mcasperson: Thanks for that feedback - I'll remove the example logic.

Steps/PitchGuidelines.md
mcasperson: 👍 I'll add these as bullet points under post release tasks.

Steps/PitchGuidelines.md
mcasperson: I think there is a lot of potential when it comes to targets, although that planning is still very much at an early stage. I'll link to some of the pitches that touch on targets in this document to give people a sense of where this might go in future.

Steps/PitchGuidelines.md
mcasperson: I do like where these discussions are leading, as targets are underutilised at the moment. Between pipelines and environment templates I think we start to see a compelling solution for complex deployment processes.

For now I'll constrain this document to details that help with designing step packages in the short term (i.e. what do I need to know when working on a pitch in 2021). But it does seem like we'll be revisiting these guidelines once we have a clearer picture on what pipelines looks like.

Steps/PitchGuidelines.md
mcasperson: That makes sense. I'll split these up into 3 documents.

PR #50, opened 2021-07-14
Proposed changes to support different types of package references

TomPeters:
As part of #project-ecs, we need to use the new step package framework in order to support package references to container images. They images should not be acquired on our workers, as their only purpose is to provide input to a cloud formation template.

Our current package infrastructure in steps from step packages does not support this well. It already has a number of shortcomings, but some of the main limitations that block this particular work are
- We can't currently support multiple packages on a step (this is just not implemented yet)
- We need to be able to constrain the feeds that are used to just container image feeds
- We need to be able to configure our deployment engine so that it doesn't try to acquire the container images on the worker

In order to support this, and problems like this going forward, this PR proposed some changes to the way we model packages in steps.

Rendered view: https://github.com/OctopusDeploy/Architecture/blob/tom/package-references/Steps/Concepts/InputsAndOutputs.md#packages


Steps/Concepts/InputsAndOutputs.md
```diff
@@ -453,6 +460,103 @@ radioButtons({
...
+
+```tsx
+interface MyInputs {
+  dockerPackage: PackageReference<DockerPackage>;
+  nugetPackage: PackageReference<NuGetPackage>;
+  dockerOrNugetPackage: PackageReference<DockerPackage | NuGetPackage>;
+}
+```
+
+Where `DockerPackage` and `NuGetPackage` are examples of the types exposed through the `step-api`.
```
TomPeters, line 513: I can see this going two ways

1. We expose a separate type for each feed type we support. Docker, nuget, maven, github, etc.
2. We expose some broad categories of packages that represent the different ways you can use these packages, eg. `ContainerImage` -> Docker + AWS ECR; `PhysicalPackage` -> Everything that can be represented as a file that can be uploaded to a worker. Are there any other categories?

Steps/Concepts/InputsAndOutputs.md
```diff
@@ -453,6 +460,103 @@ radioButtons({
...
+
+The requirement that `PackageReference`s need to map to different execution-time types based on whether they have been acquired or not means that we need to capture this information in the type system.
+
+We should have two types that represent packages that aren't acquired and packages that are acquired.
+
+```tsx
+interface MyInputs {
+  acquiredPackage: AcquiredPackageReference<NuGetPackage>;
+  notAcquiredPackage: NotAcquiredPackageReference<DockerPackage>;
+}
```
TomPeters, line 527: I can think of a few different types + naming options here

1. `AcquiredPackage` vs `NotAcquiredPackage`
2. `AcquiredPackage` vs `NotAcquiredPackage`
3. `Package` vs `PackageReference`
4. `PackageReference<"acquired">` vs `PackageReference<"not acquired">`

What is your preference?

Steps/Concepts/InputsAndOutputs.md
```diff
@@ -453,6 +460,103 @@ radioButtons({
...
+```tsx
+type ExecutionInputs = ExecutionInputs<MyInputs>; // this is equivalent to...
+
+interface ExecutionInputs {
+  acquiredPackage: { extractedToPath: string };
+  notAcquiredPackage: {
+    packageId: string;
+    packageVersion: string;
+    feed: DockerFeed;
+  }
```
TomPeters, line 541: For a step like the ECS step, we need a uri to the feed, the package id and the package version. All of these get pumped into the cloud formation template.

This representation gives us that information at execution time, plus potentially a little bit more (like feed credentials) just in case the step needs that extra information.

Steps/Concepts/InputsAndOutputs.md
shaunmarx: Just checking, did you mean to have a different option listed for `2.`? It reads the same as `1`. Did you may be mean `NonAcquiredPackage`?

TomPeters: Thanks! I fixed the original message. The second one was supposed to be package references

TomPeters:
Discussion happened in slack (https://octopusdeploy.slack.com/archives/C01GYCD6VUH/p1626241221355200) and zoom. This PR represents the outcome of that discussion.

Steps/Concepts/InputsAndOutputs.md
```diff
@@ -453,6 +460,96 @@ radioButtons({
...
+  - Acquired container images should get mapped to the same thing as non-acquired packages: the package id, its version, and information about the feed.
+
+## Solution
+
+### Assumptions
+
+- Any new steps that we write do not need to acquire container images (i.e. `docker pull`) on the workers, because we are always going to be forwarding this information to a cloud service.
+- For acquired physical packages, steps don't need to care about the source of that package (i.e. which feed type it originated from). Steps are agnostic to this detail and any step can acquire a physical package from any feed type.
+
+### `ContainerImageReference` and `PackageReference`
```
mcasperson, line 506: I think the terminology of `ContainerImageReference` will end up being too specific. [OCI artifacts](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md) are already being used to host things like [Helm charts for example](https://helm.sh/docs/topics/registries/#enabling-oci-support), and I'd expect to see Docker registries used to host a range of generic artifacts beyond container images.

TomPeters: I think that's ok. If you wanted to get a package from a docker registry that wasn't a `ContainerImageReference`, you should use the `PackageReference` type instead, which will actually acquire the package for you.

mcasperson: I'd suggest that we are already seeing deployment practices that assume the target pulls in artifacts rather than having Octopus acquire and push them. Helm deployments via GitOps is an [example today](https://github.com/fluxcd/helm-operator-get-started) - you push a declarative YAML file to a git repo that includes the reference to an external chart, and the Kubernetes cluster pulls that chart itself.

In this case Helm charts are just references like container images, and Octopus would not be responsible for acquiring and deploying them.

Projecting forward I would expect to find more examples of platforms and deployment patterns where Octopus passes around artifact references rather than directly handling the files.

Steps/Concepts/InputsAndOutputs.md
```diff
@@ -453,6 +460,96 @@ radioButtons({
...
+interface MyInputs {
+  physicalPackage: PackageReference;
+  dockerImage: ContainerImageReference;
+}
+```
+
+- The `PackageReference` type represents a selection from a feed that Octopus Server can use to download a package file. This package file is acquired on the worker.
+- The `ContainerImageReference` type represents the selection of a container image. These container images are never acquired (i.e. `docker pull`) on the workers.
+
+Steps don't need to know about which feed type was used to select or acquire these packages. This means any step that uses a `PackageRefernce` can obtain that package from any type of package feed (e.g. Built-in, NuGet, Maven, Helm, Github), while any step that uses a `ContainerImageReference` can use any container image feed type (e.g. docker, AWS ECR)
```
egorpavlikhin, line 520: `PackageRefernce` typo

Steps/Concepts/InputsAndOutputs.md
TomPeters: Let's continue this discussion in slack: https://octopusdeploy.slack.com/archives/C01GYCD6VUH/p1626312926385200?thread_ts=1626241221.355200&cid=C01GYCD6VUH

PR #51, opened 2021-09-09
Adds migration concepts for step packages

geofflamrock:
This PR adds documentation capturing discussions and decisions made for migrating inputs between versions of steps and deployment targets inside step packages.

To view the changes in a nicer preview go to https://github.com/OctopusDeploy/Architecture/blob/geoffl/step-migration/Steps/Concepts/Migration.md

# How to review

Have all the concepts and decisions that have been made as part of the planning for this feature been captured? 
Are there any edge cases that haven't been considered that can be foreseen now that this is captured?


Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+Additional complexity in migration arises when a user is attempting to update their step across multiple versions e.g. from v1 to v3. In this case there could be multiple sets of new required fields and transformations that need to be considered.
+
+## Tl;Dr
+
+- Step authors will provide migrations inside the step package to upgrade inputs for major versions only (at this stage)
+- Migrations will take current inputs and return migrated inputs
+- Users will need to review migrated inputs and save the step/target after fixing any issues
+- Multi-version migrations will run all relevant migrations and then return fully migrated inputs to the end user to be saved
+- Step authors will be expected to provide required fields and handle bound values
+- Opaque/Server types that are new in a given version will be handled specifically and will not need to be provided by the step author in a migration
```
andrewabest, line 32:
```suggestion
- Server types (like `PackageReference`) that are new in a given version will be handled specifically and will not need to be provided by the step author in a migration
```

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+- Step authors are recommended to structure their repository in a certain way to best incorporate multiple versions
```
andrewabest, line 33: Lets be more opinionated here

```suggestion
- We will use `pnpm` in step package repositories to allow multiple versions of steps to co-exist within a package and be versioned independently
```

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+- Breaking changes to the inputs of a step e.g. a new required field, or a new shape of inputs.
+- A new version of some dependent tool or cloud service that is incompatible with an existing version, even though the inputs are the same. An example of this might be a tool like Terraform where the inputs to a step may not change but the templating language has changed.
+
+Changing from one major version of a step to another will be an opt-in process for the end user configuring a step within a deployment process. As part of this change of versions, the inputs to a step may require changes to get to a new valid shape. This process of updating the saved inputs of a step/deployment target is termed _migration_.
+
+Additional complexity in migration arises when a user is attempting to update their step across multiple versions e.g. from v1 to v3. In this case there could be multiple sets of new required fields and transformations that need to be considered.
+
+## Tl;Dr
+
+- Step authors will provide migrations inside the step package to upgrade inputs for major versions only (at this stage)
```
andrewabest, line 27: Lets simplify this unless we really want to elaborate.

```suggestion
- Step authors will provide migrations inside the step package to upgrade inputs for major versions
```

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+- Migrations will take current inputs and return migrated inputs
+- Users will need to review migrated inputs and save the step/target after fixing any issues
+- Multi-version migrations will run all relevant migrations and then return fully migrated inputs to the end user to be saved
+- Step authors will be expected to provide required fields and handle bound values
+- Opaque/Server types that are new in a given version will be handled specifically and will not need to be provided by the step author in a migration
+- Step authors are recommended to structure their repository in a certain way to best incorporate multiple versions
+
+# Migration Process
+
+In order to migrate inputs between versions, a step author will be able to provide migration functions as part of the step/deployment target in the step package. These migration functions will be stored by convention into a `migrations.ts` file in the same way that the `ui`, `validation` and `executor` are stored. Each migration will store the following information:
```
andrewabest, line 37: I'm not super clear about the suggestion here, but my expectation was that any step zipped up within the step package after v1 would need to contain a single `migration.ts` file that took the vPrevious inputs and migrated them to vCurrent.

So when we selected a new version of a step, we would retrieve its migration function, and run that. 

There is a bit of an interesting implication that new steps within the monorepo are dependent on old steps, but I think the monorepo makes this less awkward than it sounds in reality.

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+  - [Common Code](#common-code)
+
+# Overview
+
+As steps and deployment targets that are created in step packages evolve over time, there will be a need for changes to a step package that result in a new major version. Reasons for these changes might include:
+
+- Breaking changes to the inputs of a step e.g. a new required field, or a new shape of inputs.
+- A new version of some dependent tool or cloud service that is incompatible with an existing version, even though the inputs are the same. An example of this might be a tool like Terraform where the inputs to a step may not change but the templating language has changed.
+
+Changing from one major version of a step to another will be an opt-in process for the end user configuring a step within a deployment process. As part of this change of versions, the inputs to a step may require changes to get to a new valid shape. This process of updating the saved inputs of a step/deployment target is termed _migration_.
```
andrewabest, line 21: Is there a way we can talk about the inputs within this document without needing to say "step/deployment target" each time? It feels like we could cut a bit of verbosity out if so.

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+- Step authors are recommended to structure their repository in a certain way to best incorporate multiple versions
+
+# Migration Process
+
+In order to migrate inputs between versions, a step author will be able to provide migration functions as part of the step/deployment target in the step package. These migration functions will be stored by convention into a `migrations.ts` file in the same way that the `ui`, `validation` and `executor` are stored. Each migration will store the following information:
+
+- A version indicating when the migration should be run. This will support a SemVer string (e.g. `2.0.0`) to allow for future flexibility in migration support.
+- A function that will take an input parameter of the current inputs and will be expected to return the migrated inputs.
+
+The migrated inputs would be set as the inputs of the step but will not be automatically saved, instead the end user performing the migration will be expected to review and adjust the inputs as they see fit and then save the updated deployment process, which will include changing the step to the new version.
```
andrewabest, line 42: I feel like we jump from step authoring to the UX here a bit awkwardly - perhaps we can have one section about authoring the migration, and another about the UX that uses the componentry?

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Required Fields
+
+Some inputs from a new version of a step will likely have fields that are required that are new and cannot be derived from the previous inputs. The migration process will require that the step author supply a value for these fields, in the same way that these fields are expected to be provided in the `createInitialInputs` function of the UI definition which is used when a new step is added. This doesn't mean that the value that is supplied as the migration default has to be a valid value, for example a new `port` field might have a migration value set to 0 as there is no way to supply a more sensible default however the valid range is >0. In this case before the user can successfully save the resultant step after migration they will need to set a valid value.
+
+In the case of multi-version migrations this lack of a valid may cause some complexity in subsequent migrations after the version that introduced the field, the step author will be expected to navigate this complexity as part of their migration scripts. This is so that the end user does not need to fill in require fields at each step of a multi-version migration.
```
andrewabest, line 56: The implication here more explicitly is that migrations can only assume incoming data is type-correct, but can make no assumptions about validity.

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Opaque/Server Types
+
+There are certain types of inputs that are opaque, where the underlying shape at configuration and execution time are controlled by Octopus Server. Examples of this include sensitive values, package/container image references and accounts. When properties of these types are part of inputs passed in to a migration function these will not be typed as the configuration/execution types. TODO: Validate this!! Could there be a situation where a migration depends on knowing something about an account or a package reference?
```
andrewabest, line 64: Migrations won't depend on knowing anything of the content of package references or accounts or any server-owned types. There is nothing within the information contained within those types a step migration should need to know about. If the opposite were true our api would have to start knowing about these server types, which doesn't feel right - if this requirement came up I'd suspect we would look at other ways of migrations discovering this information.

```suggestion
There are certain types of inputs that are opaque, where the underlying shape at configuration and execution time are owned by Octopus Server. Examples of this include sensitive values, package/container image references and accounts. When properties of these types are part of inputs passed in to a migration function these will not be typed as the configuration/execution types.
```

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+
+When new properties are added to the inputs during migration to a new version that are of one of these opaque types, the step author will not be expected to supply a value for these. These properties will be marked as optional in the output type. The migration functionality within Octopus Server will instantiate an appropriate value for these properties and input may be required from the user e.g. need to pick a package reference.
```
andrewabest, line 66: We don't really touch on the facility for this - but perhaps we don't need to. The implication is the migration output is a mapped type that will mark any server-owned types as optional.

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Minor/Patch Versions
+
+At this stage migrations will not triggered for updates to minor versions of steps, any additive properties that are optional will need to be handled appropriately by the step author within the `executor` of the step.
+
+# Step Package Repository Design
```
andrewabest, line 72: I feel like in here we need to initially talk about requiring a design that allows us to individually version each step within the repository. So start with framing the problem we need to solve, the move into the guidance on how we solve it.

andrewabest:
I've made a bunch of comments / suggestions. LMK when to take another look :)

Steps/Concepts/Migration.md
geofflamrock: That's a good thought, maybe I'll add some sort of brief glossary at the start, there are possibly other terms in here that could be clarified in the same way

Steps/Concepts/Migration.md
geofflamrock: Oh, I hadn't really considered that approach, I was imagining that the migrations would be replicated in some fashion across the various versions e.g. the v3 step would have the migrations for v1 -> v2 and v2 -> v3 (maybe through some shared code I suppose, up to the step author). The reason I thought this was so that we don't have to worry about a missing v2 package in the Server for whatever reason, but potentially this is not a concern?

Steps/Concepts/Migration.md
geofflamrock: Yep that is the implication, I'll look to make this more explicit here, thanks

Steps/Concepts/Migration.md
geofflamrock: That makes sense 👍  thanks will remove the todo

Steps/Concepts/Migration.md
geofflamrock: Hmmm yeah it's a bit unclear that the property in these cases can be required in the inputs that the ui etc. uses but the output of the migration for that same property is optional, I wonder if an example would be useful to clarify? Maybe just the shape of the inputs used elsewhere in the api and what the equivalent shape of the output of the migration looks like.

Steps/Concepts/Migration.md
geofflamrock: I ended up linking off to the Inputs documentation and making the word italic like we do for other key terms, and then removed some verbose text, hopefully this makes it a bit nicer to read

Steps/Concepts/Migration.md
geofflamrock: I've split into two sections for authoring and process

Steps/Concepts/Migration.md
geofflamrock: Have made this more explicit

Steps/Concepts/Migration.md
geofflamrock: I've made it more explicit that that the inputs themselves don't change, but that the migration api will provide the functionality automatically. The documentation in `InputsAndOutputs.md` does a good job of explaining this so have linked off to that too.

Steps/Concepts/Migration.md
geofflamrock: Have added some detail about the problem and made the recommendations a bit more explicit, hopefully it reads a bit better

geofflamrock:
@andrewabest thanks for the feedback 👍 I've made changes that address most of the comments you've made. There is the question still remaining about having a single migration per step version or multiple which we may need more discussion on.

Steps/Concepts/Migration.md
andrewabest: > The reason I thought this was so that we don't have to worry about a missing v2 package in the Server for whatever reason, but potentially this is not a concern?

This is a fair point. For now Server is going to be "greedy" when acquiring step packages - it will always have all available ones if it has any of them. But for offline scenarios people may be able to add them independently.

One option could be that the step package always contains the "latest version" of all major versions.

If we did pursue an approach were every migration had to cover the gap between v1 and vCurrent, how could we enforce that so that step authors were clear on what they would need to do?

geofflamrock: > If we did pursue an approach were every migration had to cover the gap between v1 and vCurrent, how could we enforce that so that step authors were clear on what they would need to do?

There was talk of adding a build time check that migrations had been supplied for prior versions over [here](https://github.com/OctopusDeploy/Architecture/blob/main/Steps/Concepts/Versioning.md#versioning):

"We may enforce this consistency within the Step Package CLI - we could refuse to build a step of version v(n) if the Step Package did not present n-1 upgrade functions".

This idea could be extended to enforce that all relevant migrations are present when the packages are built? We might need to consider deprecation as well, potentially removed versions are no longer required to have migrations, although that might cause issues for a user trying to get off a removed version of a step.

An assumption here would be that individual versioning of step packages within the build process are already in place so we could detect which versions are required to have migrations.

I really like the `exhaustiveCheck` concept that we use in the portal when adding new UI components, I wonder if we could incorporate something in that to make it really obvious where any missing migrations might need to be added?

andrewabest: Thinking about this more, I think we need to establish some constraints for this model. Here are a few off the top of my head - can you think of any more that would be useful?

Constraints:

- It should be impossible for Server to have a set of steps / packages that leads to "impossible scenarios"
	- example: Missing migrations - Have v1 and v3 of a given step, don't have the v2 step - can't upgrade from 1 to 3 as missing a migration
	- example: Processes or targets exist that reference step versions Server cannot resolve (server should not be able to "lose" a step, or create a step from a package it does not have reference to)
- Creating a new major version of a step should be a simple and safe process for a step author
	- type safety should give strong guarantees
- Migrations must be testable

I think to support the first constraint, we should ensure a step package always contains all major versions of a step. Azure devops does this currently with their tasks - the terraform task has both v1 and v2 in the package. This will give us strong guarantees that we always have a viable migration pathway, and avoids needing to "accumulate" migrations between versions. This also supports the second constraint - this model is simple to reason about for authors - each migration is always v-1 to v, nothing more. 

Testing for migrations is also a must, and should be straight-forward given we are talking about feeding concrete types in instead of input paths - we would just need to solve the "testing server-owned types" problem.

geofflamrock: I can't really think of any more constraints in this space outside of the ones you've mentioned.

It feels like there might be some implications of these constraints for future work such as discovery and acquisition of packages which might be worth considering a bit before we lock these in.

For example:
- If v1 of a step was deprecated and then removed, then the new package that no longer contains v1 wouldn't be able to be installed into Server if there are steps referencing that version, and these steps would need to be migrated before the new package could be installed. If there aren't any steps referencing v1 then installation would be ok I expect. That might lead us down the path of not supporting removal at all, only deprecation.
- If the latest minor version of a step has been marked as "poisoned" and therefore is no longer available to be used, but there is a step referencing that version (where the issue was found), how does this get resolved such that the "no impossible scenarios" constraint is satisfied. Again this could lead us down the path of not supporting that kind of "poisoned package" however this would lead to other negative side effects.

Some other thoughts:

> we should ensure a step package always contains all major versions of a step

This feels like a good decision, hopefully it leads to a more obvious and easier to understand model for authors.

> type safety should give strong guarantees

This will definitely help us a lot, I suppose an author might use the wrong type associated with a version for a migration though accidentally e.g. migration function for v2 -> v3 takes in `MyInputsV1` as the inputs instead of `MyInputsV2`. I wonder how far we would expect the API to go to protect against this sort of thing and where we draw the line at saying the author of the step just needs to do a good job. 

> Testing for migrations is also a must, and should be straight-forward given we are talking about feeding concrete types in instead of input paths - we would just need to solve the "testing server-owned types" problem.

I wonder if not requiring the author to set a value for any new properties of server owned types will help here, most migrations involving existing properties of these types I imagine would be relatively simple e.g. move the location of my property from one spot to another because of a new shape.

andrewabest: Alright I think we are mostly in agreement. I think deprecation without removal is a reasonable starting point ot assume. And the complexity around poisoning packages sounds manageable. Can you update migration section of the PR to reflect this discussion, then I'll give it one last pass over before we ship it?

geofflamrock: @andrewabest the Authoring migrations section is updated in line with what we've discussed, and I've added a section outlining the constraints as well. 

One last question remains for me on what I've just added, should migrations for a given version be optional? If so I've noted that we would assume that the shape of the inputs match for the new version and will just continue on with migration. Happy to change it to force migrations to be provided if we think that is better.

andrewabest: @geofflamrock I think the safety mechanisms will be a lot easier to implement if they are mandatory - it should be simple to write a no-op migration anyway and just return the inputs.

geofflamrock: Yep cool, I'll change that bit, thanks 👍

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,144 @@
...
+- We will use `pnpm` in step package repositories to allow multiple versions of steps to co-exist within a package and be versioned independently
+
+# Constraints
+
+The following constraints will apply when considering migrations:
+
+- It should be impossible for Octopus Server to have a set of steps or packages that result in "impossible scenarios". Some examples of impossible scenarios:
+  - Missing migrations: Server has v1 and v3 of a given step as a package installed didn't contain v2 and therefore there is no way to migrate a step from v1 -> v3.
+  - Missing packages: Deployment processes or deployment targets exist within Server that reference versions of packages that cannot be resolved e.g a deployment process references v2 of a step but that version doesn't exist in Server anymore.
+- Create a new major version of a step should be a simple and safe process for a step author
```
andrewabest, line 46:
```suggestion
- Creating a new major version of a step should be a simple and safe process for a step author
```

geofflamrock: Fixed 👍

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,142 @@
...
+        |-- validation.ts
+        |-- ...
+    |
+    |-- my-step-v2
+        |
+        |-- executor.ts
+        |-- inputs.ts
+        |-- ui.ts
+        |-- validation.ts
+        |-- migrations.ts
```
andrewabest, line 122:
```suggestion
        |-- migration.ts
```

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,142 @@
...
+        |-- ...
+```
+
+To create a new version of a step in the step package repository:
+
+- Create a new folder representing the new version of the step/target
+- Copy any relevant contents from the previous step e.g. inputs, ui, validations, executor
+- Adjust contents as required. Particularly adding a version number to the inputs will aid in readability for migration functions etc. e.g. use naming like `type MyStepInputsV1` and `type MyStepInputsV2`
+- Add migration functions to migrate from the previous version of the step/target to the new version
```
andrewabest, line 131:
```suggestion
- Add a migration function to migrate from the previous version of the step/target to the new version
```

andrewabest (approved):
Looks good to circulate wider now 👍

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,142 @@
...
+  - Missing packages: Deployment processes or deployment targets exist within Server that reference versions of packages that cannot be resolved e.g a deployment process references v2 of a step but that version doesn't exist in Server anymore.
+- Creating a new major version of a step should be a simple and safe process for a step author
+- Migrations must be testable
+
+# Authoring Migrations
+
+In order to migrate inputs between versions, a step author will be expected to provide a migration function as part of the step/deployment target in the step package which will take in inputs from the previous version and will be expected to return the migrated inputs matching the current version.
+
+In order to satisfy "No impossible scenarios" constraint mentioned above, all _major_ versions of a step/target must always be contained within a step package repository (and published step package), and therefore each migration function will only need to consider the inputs from the previous version. All versions of a step/target above v1 will require a migration, even if it is just returning the inputs unmodified (in the case of a migration where nothing has changed).
```
TomPeters, line 53: > all _major_ versions of a step/target must always be contained within a step package repository

Let me challenge this one. 

Using the dbup style migration model as inspiration here, what would it look like if the step package _only_ contained the implementation for the current step version, and we had a migration function that knew how to migrate from older step versions up to the current version?

Thinking about the "Missing Migrations" and "Missing packages" scenarios: The latest version of the step contains all of the migrations from v1 of the step, just like how DbUp starts with the v1 schema.

Any version of a step is probably in hotfix territory, and we would only be doing real maintenance on the latest version of a step (is this claim true?), so we would rarely need to modify older versions of a step. If we do, we can go to the _branch_ for that step version, make the change and publish a new version, and merge that change forward to future versions of the step.

One advantages of this approach is that we can iimplify the tooling: we don't need pnpm or changesets.

Another advantage is that we don't need to maintain that older (deprecated) versions of the steps are compatible with the current step api. Imagine we changed the `step-ui` interface to add another required property, or a breaking change to a component. With the solution being proposed in this PR, we would need to update every historical version of a step, even if those steps aren't being used anymore. If we only include the latest version of a step, we don't need to do this pointless work.

What do you think?

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,142 @@
...
+
+In the case of multi-version migrations the implication for the step author of required fields that may not be valid is that the types of the inputs to a migration function will be correct but no assumption can be made about the validity of data. This may cause some complexity in subsequent migrations after the version that introduced the field, the step author will be expected to navigate this complexity as part of their migration scripts. This is so that the end user does not need to fill in require fields at each step of a multi-version migration.
```
TomPeters, line 61: > In the case of multi-version migrations the implication for the step author of required fields that may not be valid is that the types of the inputs to a migration function will be correct but no assumption can be made about the validity of data.

Let me play devil's advocate here.

What would it look like if we wanted to enforce you could only do the next step of a migration if the inputs were valid? For example, if you had migrated from v1 -> v2, and then the inputs were invalid, the migration could not proceed from v2 -> v3.

A design philosophy that we should be adopting is to make our steps as simple as possible to configure by giving them sane defaults and having as few as possible required fields.

Thinking about some of our simple steps like deploy a package or run a script, they have about 1 field you need to provide to make them valid.

Based on this, how often would we add a new required field that couldn't be made valid by default? Said another way: What is the likelihood that we would encounter a migration that wouldn't leave things in a valid state? 

I think it would be unlikely, but if it does arise what would you do about it? If you had access to the intermediate step, you might show the UI and ask the user to fix the error before continuing.

I think the complexity of this workflow for a rare edge case probably doesn't make building such a feature worthwhile, and I think the proposal in this PR (assume the incoming data might be invalid) is the way to go :+1:

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,142 @@
...
+
+The migration functionality within Octopus Server will instantiate an appropriate value for these properties and input may be required from the user e.g. need to pick a package reference.
+
+# Migration Process
+
+The migration process will be initiated by the end user when configuring a step that has an available upgrade to a newer major version. After the migration process is completed (running the relevant migration functions outlined in [Authoring Migrations](#authoring-migrations)) the migrated inputs will be set as the inputs to the step and the updated UI shown to the user. The step will not be automatically saved, instead the end user performing the migration will be expected to review and adjust the inputs as they see fit and then save the updated deployment process, which will include changing the step to the new version.
```
TomPeters, line 77: Does this migration occur on the backend or frontend? Probably worth being explicit about this.

Steps/Concepts/Migration.md
```diff
@@ -0,0 +1,142 @@
...
+
+In the case of migrating across multiple major versions each migration function will be run in order as part of the process and the output of one migration function would become the inputs to the next migration function. Only once all relevant migrations have been run would the results be returned to the user for saving. For example upgrading from v1 to v3 would involve:
+
+- Running the v1 -> v2 migration with the current inputs
+- Running the v2 -> v3 migration with output of the v2 migration
+- Returning the results of the v3 migration and setting these as the inputs of the step/target before user review and save
+
+## Minor/Patch Versions
+
+At this stage migrations will not triggered for updates to minor versions of steps, any additive properties that are optional will need to be handled appropriately by the step author within the `executor` of the step.
```
TomPeters, line 89: I can imagine wanting to add a new optional property as part of a minor version bump, and wanting to give it a default value that isn't `undefined`, just like you can with `createInitialInputs`.

Does this mean we should run migrations for minor version bumps too?

TomPeters (commented):
A few comments and thought provoking questions to consider :)

Steps/Concepts/Migration.md
gb-8: We talked about migrations for minor versions, and the conclusions we reached were that given (a) we want Octopus to automatically use the most recent version with matching major version, and (b) migrations are currently planned to be executed at configuration time, then we felt that having step authors allow for newly added optional properties to be undefined in the execution script (and fall back on a default value there) was preferable to executing migrations for minor version bumps at execution time.

gb-8:
Should we be updating the existing Versioning documentation as part of this PR to reflect where we have arrived at now? And maybe add some more information there about how versioning would need to work based on the new migrations plan?

Steps/Concepts/Migration.md
geofflamrock: We did discuss that type of approach (trying to be "smart" about when the user needed to provide input along the way to ensure that inputs are valid at every stage of a migration for the step author) and came to the same conclusion that you did, the cases where adding a new required field where the default is invalid in subsequent versions of a step are fairly unlikely, and the additional complexity wasn't worth it. I appreciate the thought process you've outlined here, helps to validate what we were thinking along the same lines 👍 .

geofflamrock:
> Should we be updating the existing Versioning documentation as part of this PR to reflect where we have arrived at now? And maybe add some more information there about how versioning would need to work based on the new migrations plan?

Yes we may want to clean up a little bit of the versioning documentation, mostly there are a few descriptions of what we _might_ want to do with regards to migration functions and enforcing in the cli. I'll take a look and adjust, thanks.

Steps/Concepts/Migration.md
geofflamrock: It's intended to be run on the server, so that we can control the way the migration happens and so that other non-portal clients can upgrade steps too. I'll make that more explicit, thanks.

Steps/Concepts/Migration.md
andrewabest: Musing out alout here regarding a dbup-style approach:

- you would still need to accrue input versions - the migration function needs to know what the expected input shape is, and the output shape. Perhaps the old input shapes would just be part of the migration api. This doesn't feel like a big deal.

- If we want to actively develop multiple major versions at once, say to keep feature parity across them, it feels like the versions coexisting is a better model - an example might be terraform, our our existing Azure Webapp step if people want to use webdeploy instead of zip deployments. 

- the step package size itself would be smaller with this approach as it wouldn't need to contain all versions

- it introduces the need for an LTS-branch style approach, which feels like it goes against the simplicity constraint.

If we use a "feature branch hotfix" model, and we have multiple steps and targets existing in a step packages as we do at the moment, how do we branch and tag to enable independently shipping hotfixes to any of the contents? It is probably a solveable problem, but it feels complex to reason about. 

I like the simplicity of reasoning bought about by having all versions present in the repository at any given time. 

> One advantages of this approach is that we can iimplify the tooling: we don't need pnpm or changesets.

We are using changesets for release notes regardless of monorepo status, so this is partly true.

> With the solution being proposed in this PR, we would need to update every historical version of a step, even if those steps aren't being used anymore

That isn't true, there are a few ways to solve this problem. We are using pnpm to allow independent versioning and publishing, but you could also just have seperate steps in a repo without the monorepo tooling. Both would allow each version to have a different set of dependencies.

geofflamrock: With multiple steps and targets in the same repo with the desire for individual versioning, it would make any branching strategy more complex.

Some potential approaches I've been thinking about:

_Keeping major versions in lockstep and branching into `release/vPrevious`_ 

e.g. when a v2 of a step needs to occur then all the steps/targets get bumped and we maintain a `release/v1` branch for hotfixes. 

This would probably only really work if there were a limited number of steps/targets that were quite closely connected and it made sense to bump them at the same time. There would likely also be lots of unnecessary upgrades of some steps which doesn't feel great. 

This feels similar to the way the Calamari branching works where we have a branch associated with a server release and everything gets major bumped.

_Creating a specific branch for `vPrevious` of a specific step when required._ 

When `vNext` of a step is created a particularly branch could be created for the `vPrevious` of the step e.g. `release/azure-blob-storage-v1`. Each step would have its own set of branches as time went on, the main branch would continue to have the `vCurrent`.

This has some downsides of potentially lots of branches building up over time (prob not a huge deal), and each branch would also contain whatever version of the other steps were at the time, this may not be a big deal if it was easy to know which branch needed to be hotfixed. A good naming convention might help perhaps. 

There's also the complexity of the connection between steps and targets to consider in this model, if there is a change that needs to be made to the target does that need to happen in all the LTS branches for each step that is connected to that target? Potentially not depending on the change in question and if we continue to treat target versions as essentially "fixed" then it probably also becomes less of an issue.

---

One other thing I've been thinking about with this is that the complexity is mainly due to the fact that we have multiple steps within the same repo/package that may not really be connected, I wonder if limiting to a single step+target would make things simpler. This may not be desirable for other reasons though.

geofflamrock: We've been discussing this further and have decided to stick with the current approach of storing major versions side-by-side within the same repo. We're looking to design the migration api that a step author would use such that we could change this approach later on without requiring the author to update to a newer api etc. We'll also spin up a spike of what this approach might look like in an example step package repo.

We'll also capture this alternative idea in a new section of this document so we have the context.

Thanks @TomPeters for raising this and prompting the discussion :)

PR #53, opened 2021-09-29
Updates step package migration documentation

andrewabest:
This PR updates our step package documentation to reflect recent decisions made regarding how migrations will be defined, and how packages will be structured. 

- Each step will provide the set of migrations required to take it from v1 to vCurrent
- Step packages will be independently versioned, with each package only containing a single "thing" (step or target)

Context: https://octopusdeploy.slack.com/archives/C02D9DHAE78/p1632871549200900


Steps/Concepts/Migration.md
```diff
@@ -17,86 +17,108 @@
...
 When new properties are added to the inputs during migration to a new version that are of one of these opaque types, the step author will not be expected to supply a value for these. These properties will be marked as optional in the output type automatically by the migration api, however the property in the inputs that get passed to the `ui` and `executor` will still be defined as required. See [Bounded Contexts](./InputsAndOutpus.md#bounded-contexts) for more detail on this approach.
 
 The migration functionality within Octopus Server will instantiate an appropriate value for these properties and input may be required from the user e.g. need to pick a package reference.
 
-# Migration Process
+# User Experience
+
+The migration process will be initiated by the end user when configuring a step that has an available newer major version.
+
+The migration process will be run by Octopus Server, which will run the relevant migration functions outlined in [Authoring Migrations](#authoring-migrations) and upon completion return the migrated inputs, which will be displayed in an updated step UI to the user.
```
geofflamrock, line 86:
```suggestion
The migration process will be run by Octopus Server, which will run the relevant migration functions outlined in [Authoring](#authoring) and upon completion return the migrated inputs, which will be displayed in an updated step UI to the user.
```

Update to match the change of heading above

Steps/Concepts/Migration.md
```diff
@@ -17,86 +17,108 @@
...
+- It should be impossible for Octopus Server to have a set of step packages that result in "impossible scenarios". Some examples of impossible scenarios:
+  - Missing migrations: Server has v1 and v3 of a given step package but not v2. Just because server is missing an intermediary version does not mean it should to be able to perform the upgrade from v1 to v3.
+  - Missing packages: Deployment processes or deployment targets exist within Server that reference versions of step packages that cannot be resolved by Server
+- Creating and maintaining new major versions of a step package should be a simple and safe process for a step package author
 - Migrations must be testable
 
-# Authoring Migrations
+# Authoring
+
+For each major step package version > 1, a step author will be expected to provide the set of migration functions required to take the inputs from `v1` to `vCurrent`. This satisfies the _no impossible scenarios_ constraint. How this is achieved is left to the step package author, as long as it fits the defined migration API shape.
```
geofflamrock, line 56: I wonder if its worth being explicit here about the fact that we expect the set of sequential migration functions (e.g. v1 -> v2 and v2 -> v3), not some other set of combinations?

geofflamrock: It is captured down in the Multi-Version Migrations section so may not need to here

Steps/Concepts/Migration.md
```diff
@@ -17,86 +17,108 @@
...
+
+To make supporting concurrent versions of step packages simpler, we recommend a repository design that has each supported version of a step package co-existing side-by-side, rather than relying on git branches to support this co-existence.
 
-The repository for step packages needs to be designed to meet these requirements.
+The goal of this structure is to make it simpler to add and maintain concurrent versions of steps over time, which helps satisfy our _Creating and maintaining new major versions of a step package should be a simple and safe process for a step package author_ constraint.
 
 ## Recommendations
 
-It is recommended that multiple versions of steps/targets be stored side-by-side in a monorepo for the lifetime of versions. This will allow for maintenance of older version as well as publication of multiple versions of a given step/target within the same package for consumption in Octopus Server.
+It is recommended that multiple versions of step packages be stored side-by-side in a monorepo for the lifetime of each versions.
```
geofflamrock, line 114:
```suggestion
It is recommended that multiple versions of step packages be stored side-by-side in a monorepo for the lifetime of each version.
```

Steps/Concepts/Versioning.md
```diff
@@ -20,37 +20,34 @@ _Compatibility_ refers to the compatibility of various Step Package components w
...
+We have a step like `step-package-azurestorage.1.0.0` which is compiled against `step-ui-api.1.0.0`. We then make a breaking change to the `step-ui-api` and release a new version `step-ui-api.2.0.0`./and update our framework in portal as well, so that `step-package-azurestorage.1.0.0` is no longer directly compatible with portal. We can either upgrade the `step-ui-api` dependency and release a new version of our step (say `step-package-azurestorage.2.0.0`) to make it compatible and deprecate the old step somehow, or we could have a compatibility shim in portal so that portal can support steps compiled against both `step-ui-api.1.0.0` and `step-ui-api.2.0.0`.
```
geofflamrock, line 36: Was the `./` expected here?

Steps/Concepts/Versioning.md
```diff
@@ -20,37 +20,34 @@ _Compatibility_ refers to the compatibility of various Step Package components w
...
 
 **Minor Version Increments:** Within Server, we will automatically retrieve the _latest version_ of the same major version a step was created with when working within any of the steps components - UI, validator, executor, etc. This means for all backwards-compatible feature additions and bug fixes, users will recieve these benefits immediately in their existing processes, without having to update anything.
 
-**Major Version Increments:** For major version increments, a user will need to opt-in to them. There will be a version selector presented on the step UI if a newer version is available. If it is, and the user selects the next version, the step will be upgraded to the selected version. This will not be a reversable process.
+**Major Version Increments:** For major version increments, a user will need to opt-in to them. There will be a version selector presented on the step UI if a newer version is available. If it is, and the user selects the next version, the step will be upgraded to the selected version
```
geofflamrock, line 48: It says we will show a version selector to the user, but in the typical upgrade path we won't show a version selector, it'll just be a button, this could potentially change to reflect that.

geofflamrock (approved):
Changes look good and match up with what we discussed, I made a few minor suggestions and had a couple of questions but looks good 👍

Steps/Concepts/Versioning.md
andrewabest: Sure, I could probably make this more generic - it doesn't really matter what the UI is.

Steps/Concepts/Versioning.md
andrewabest:
```suggestion
We have a step like `step-package-azurestorage.1.0.0` which is compiled against `step-ui-api.1.0.0`. We then make a breaking change to the `step-ui-api` and release a new version `step-ui-api.2.0.0` and update our framework in portal as well, so that `step-package-azurestorage.1.0.0` is no longer directly compatible with portal. We can either upgrade the `step-ui-api` dependency and release a new version of our step (say `step-package-azurestorage.2.0.0`) to make it compatible and deprecate the old step somehow, or we could have a compatibility shim in portal so that portal can support steps compiled against both `step-ui-api.1.0.0` and `step-ui-api.2.0.0`.
```

Steps/Concepts/Versioning.md
andrewabest:
```suggestion
**Major Version Increments:** For major version increments, a user will need to opt-in to them. There will be a visual indication presented on the step UI if a newer version is available. If it is, and the user opts into the next version available, the step will be upgraded to the newer version
```

PR #57, opened 2021-11-09
Git Schema Migrations

YuKitsune:
❤️ 📄 Friendly view: https://github.com/OctopusDeploy/Architecture/blob/eoin/cac/document-migrations/OctopusServer/ConfigAsCode/Migrations/Migrations.md


zentron:
This is well summarized  nice work👍  
Will be interesting to get someone else on the team to read through it and make sure it follows.

acodrington:
Thank you for the extra explanations and diagrams at the beginning! They are a big help in understanding this very tricksy feature. Great job!

I think a newbie will probably still have to read the article over a few times before it sinks in (at least, I did), but that's because it is actually complicated. This is a v. good summary.

PR #58, opened 2021-11-22
Add mapping guidelines.

uglybugger:
Adds mapping guidelines to start moving us away from the old `IResourceMapper` pattern.


OctopusServer/Mapping/Mapping.md
```diff
@@ -0,0 +1,52 @@
...
+
+## FAQ
+
+### What about convention-based mapping? Isn't that faster and easier?
+
+Convention-based mapping makes the assumption that the source and target are very similar. This is generally not a good assumption as it misses the point of mapping. It is also fragile and prone to breaking in unexpected ways. It is entirely reasonable for a developer in the core of the system to rename a property of something internal, and this renaming should not result in a runtime mapping failure.
+
+Reflection-based mapping is unarguably slower. Dynamic, reflection-based code emission can generally perform on par with hand-written code but the price is developer confusion and fragility.
+
+Yes, we have heard of AutoMapper. Please do not use AutoMapper.
```
gupta-kartik, line 22: As I read this, I questioned, "Why should I not use AutoMapper?"

OctopusServer/Mapping/Mapping.md
```diff
@@ -0,0 +1,52 @@
+# Mapping
+
+Mapping in Octopus Server is achieved via two key interfaces: `IMapToNew<TSource, TTarget>` and `IMapToExisting<TSource, TTarget>`.
+
+* `IMapToNew<TSource, TTarget>` takes an instance of `TSource` and returns a new instance of `TTarget`.
+* `IMapToExisting<TSource, TTarget>` takes instances of both `TSource` and `TTarget` and maps the former onto the latter.
+
+Both of these interfaces are intentionally async-only.
+
+When creating a mapper, please name it `Map<SourceType>To<TargetType>`, e.g. `MapFromFooToFooResource` or `MapFromCreateDeploymentCommandV1ToCreateDeploymentCommandV2`. Some mappers will implement both `IMapToNew<,>` and `IMapToExisting<,>` and this is generally fine.
```
MarkSiedle, line 10:
```suggestion
When creating a mapper, please name it `MapFrom<SourceType>To<TargetType>`, e.g. `MapFromFooToFooResource` or `MapFromCreateDeploymentCommandV1ToCreateDeploymentCommandV2`. Some mappers will implement both `IMapToNew<,>` and `IMapToExisting<,>` and this is generally fine.
```

OctopusServer/Mapping/Mapping.md
```diff
@@ -0,0 +1,52 @@
...
+
+If mapping onto an entity, the entity itself is responsible for refusing to be put into an invalid state. Bolt-on validation is undesirable here.
+
+## History
+
+Mapping within Octopus Server has historically been closely coupled to the concepts of a `Model` and a `Resource`. This is a coupling which we need to break in order to move forward.
+
+You will find quite a number of locations in the codebase where this still the case. Please do not perpetuate this - the interfaces we would like to encourage everyone to use are simply `IMapToNew<TSource, TTarget>` and `IMapToExisting<TSource, TTarget>`.
+
+As Octopus Server moves progressively further away from a RESTful API, mapping will become correspondingly less symmetrical. Mapping from a domain entity to a DTO will generally be a "copy the relevant properties" affair, whereas mapping from a command (or a legacy `POST <resourceDto>`) to a domain entity will be a series of method calls onto that entity, giving that entity the discretion about whether to accept the caller's changes.
```
gupta-kartik, line 52: Given this repo is public, are you happy for this info to be in the public domain at this time?

gupta-kartik (approved):
LGTM. Thank you for writing this up. 👍🏽

OctopusServer/Mapping/Mapping.md
```diff
@@ -0,0 +1,52 @@
...
+
+## FAQ
+
+### What about convention-based mapping? Isn't that faster and easier?
+
+Convention-based mapping makes the assumption that the source and target are very similar. This is generally not a good assumption as it misses the point of mapping. It is also fragile and prone to breaking in unexpected ways. It is entirely reasonable for a developer in the core of the system to rename a property of something internal, and this renaming should not result in a runtime mapping failure.
+
+Reflection-based mapping is unarguably slower. Dynamic, reflection-based code emission can generally perform on par with hand-written code. The price for either convention-based approach, however, is developer confusion and fragility.
+
+Yes, we have heard of AutoMapper. We do not intend to use AutoMapper.
```
MarkSiedle, line 22: +1. Seems like a good idea when things are small (_"Hey look how fast I can map all-the-things 🎉"_, but (at least in my experience with it) can lead to some unforeseen negatives for DX (being able to find assignments is impossible, debugging, refactoring issues can quickly become difficult to reason about and leads to more tech debt than you'd expect).

Being explicit about our mappings seems onerous initially, but comes with a lot of advantages, mostly around clarity, simplicity and (we believe) is more sustainable long-term.

MarkSiedle (approved):
Thank you 🙇 This is really useful to see where we're heading for solving mapping complexities and tech debt that's hurting us now.

OctopusServer/Mapping/Mapping.md
uglybugger: Agreed. And by the time that realisation occurs, it generally occurs because it's become a large, expensive mess to fix.

Once we have a little more robustness around our DTO convention tests, we should also add rules for constructor parameters which will allow us to move away from the object initialiser pattern to just a straight "create the target via a parameterised constructor call" approach, which will then give us compile-time errors if we don't map a required property.

PR #59, opened 2021-11-23
Update API guidelines according to #project-api-philosophy

uglybugger:


OctopusServer/Api/ApiPhilosophy.md
```diff
@@ -0,0 +1,463 @@
...
+
+There are a number of reserved top-level HTTP route namespaces.
+
+```
+/app            // Reserved for the Octopus portal (web UI).
+/api            // Reserved for Octopus Server core components.
+/extensions     // Reserved for dynamically-loaded extensions.
+```
+
+When an extension is registered, its HTTP routes (advertised via MVC controller `[Route("...")]` annotations) will be asserted to start with its extension name. E.g. when an extension named Octopus.Extensions.Slack is loaded, it will have the namespace `/extensions/octopus.extensions.slack` reserved for it. If it attemps to register an MVC route not staring with that prefix then the extension will not be permitted to load.
```
paulegradie, line 300: noticed a missing T while reading 'staring' 🙏

OctopusServer/Api/ApiPhilosophy.md
```diff
@@ -0,0 +1,463 @@
...
+var command = new ConfigureSlackNotificationCommandV1(...);
+await client.Do(command, cancellationToken);
+```
+
+and the same `Octopus.Client` library would be able to work out how to deliver the payload to the correct HTTP endpoint using the correct verb.
+
+Likewise, a deployment Bash/PowerShell script step would be able to send a service message like this:
+
+```
+##octopus[SendSlackNotificationCommandV1 channel='#team-core-platform' message='All your build are belong to us! Ha ha ha!']
```
paulegradie, line 460: Haha nice [reference](https://www.youtube.com/watch?v=jQE66WA2s-A)

OctopusServer/Api/ApiPhilosophy.md
uglybugger: Well spotted. Thank you 👍

OctopusServer/Api/ApiPhilosophy.md
```diff
@@ -0,0 +1,463 @@
...
+1. `com.octopus.client.jar` (HTTP-aware client; does not know about the other two packages.)
+
+and so on.
+
+Importantly, `Octopus.Client` does not know about either `Octopus.Server.MessageContracts.HttpRoutes` or `Octopus.Server.MessageContracts` - it just knows about the generic concepts of `ICommand<,>`, `IRequest<,>` etc. It also knows about the `IHttpRouteFor<T>` interface, where `T` is either a request or a command. In other words, the client is simply given a payload and can figure out where to send it and which HTTP verb to use.
+
+A sample use case might look like this:
+
+```csharp
+var request = new GetDeploymentRequestV4(...);
```
andrewabest, line 389: How do we solve the problem of developer experience with this model?

A natural thing for a developer to do when using a client would be to find a "repository" or "client" for the thing they want to work with - for example a project client. They would create the client `var client = new ProjectClient(...)`, and then explore what capabilities for working with projects that client would provide via intellisense `client.{...}`.

With the proposed model, developers would have to hunt for classes instead - how do we provide cognitive boundaries / a path to success for developers who want to work with these clients?

andrewabest:
A lot of these proposed changes make tradeoffs that support our technical direction for Octopus, but come at some cost to customer experience with the API or clients as a part of our product. 

Some examples include:

- Asymmetric resources: these allow us to write more specific code and to rely on our abstractions, but make the client harder to work with (can't just GET a thing, modify it, then PUT it back)
- Thin client: this allows us to provide specific versioned resources for commands or queries, but detract from developer experience as locating the contracts is more difficult without clients that provide contextual boundaries (i.e. `projectClient`)
- Routes: having routes inferred from contracts gets away from hypermedia and supports people building clients against the Octopus OpenAPI spec, but the tradeoff here is customers working with our API have to understand how to establish routes for the clients they build against our Octoups OpenAPI spec.
- Command / Query based API: having a cqrs-based API surface maps to how we want to work with our API as developers, but the tradeoff is that generating a client from the OpenAPI surface of a CQRS-based API would likely end up producing a client that would be largely unuseable for a developer.

One way we might address paying for these costs would be to develop a SDK on top of the proposed API and client that provided a more developer-friendly experience to drive Octopus with.

PR #63, opened 2022-01-05
Octopus HQ Integration

pawelpabich:
Based on https://docs.google.com/document/d/1wBETbT73xgnFvrxHIQogh-XR3_FNG8981lMs9Gk_4-c/edit#heading=h.pd5a0x9y19sg

Events will be defined as a set of JSON Schema files


OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,93 @@
...
+    "specVersion": "1",
+    "payload": { }
+}
+```
+
+| Name        | Type        | Description | Implementation |
+| ----------- | ----------- | ----------- | ----------- |
+| id          | string      | The unique indentifier | This property can be used to take advantage of [the message deduplication feature offered by Azure Service Bus](https://docs.microsoft.com/en-us/azure/service-bus-messaging/duplicate-detection). At the minimum the identifier needs to be unique within the topic it is used.
+ |
+| source   | string        | The name of the source system | |
```
pawelpabich, line 87: I dropped `path` because I don't think it is being used

OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,93 @@
...
+| type   | string        | The type of the event | |
+| occurredTimeUtc   | string        | An ISO 8601 formatted string representing the data and time of when the event occured  | |
+| activityId   | string        | An identifier that identifies the overall activity in traces | |
```
pawelpabich, line 90: We need to add some additional info that makes sure the distributed tracing works. I will follow up in the slack channel

OctopusHQ/Events.md
```diff
@@ -0,0 +1,26 @@
...
+
+| Property | Type        | Description | 
+| ---------| ----------- | ----------- |
+| CloudSubscriptionId   | string(200)  ||
+| DnsPrefix   | string(4-100)  ||
+| OctopusIdClientId  | string(50)  ||
+| OctopusIdClientSecret  | string(200)  |Base64 encrypted value|
+| RegionName   | string(50)  ||
+| RetentionLimitInDays   | int?  ||
+| TaskCap   | int(5\|10\|20\|40\|80\|160)  ||
```
flin-8, line 15: The number/values inside the ()'s sometimes mean "length" and sometimes "allowed values". Can we make it more explicit, like:

| Property | Type        | Description | 
| ---------| ----------- | ----------- |
| CloudSubscriptionId   | string(200)  ||
| DnsPrefix   | string(4-100)  ||
| OctopusIdClientId  | string(50)  ||
| OctopusIdClientSecret  | string(200)  |Base64 encrypted value|
| RegionName   | string(50)  ||
| RetentionLimitInDays   | int?  ||
| TaskCap   | **int <br/>Allowed values: [5\|10\|20\|40\|80\|160]**  ||

OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,99 @@
...
+
+A good event has the following properties:
+
+* It’s treated as a strict contract that needs to be honoured for at least 6 months which translates currently to 2 3month longs cycles/quarters. In other words, it is unlikely to change.
+* It's frequency can be estimated reasonably well. E.g. 10s/100s/1000s a minute/an hour.
+* It’s specific (e.g. `SubscriptionTerminatedV1Event`, `OutageWindowChangedV1Event`) not generic (e.g. `SubscriptionUpdatedV1Event`)
+* Its entire payload is mandatory. E.g. All attributes have values assigned to them.
+* It only includes data specific to what it represents (e.g. `OutageWindowChangedV1Event` shouldn’t carry subscription limits just in case one of the consumers needs them)
+* It is published only when a real change has occurred (e.g. `OutageWindowChangedV1Event` should not be published if `Start` and `End` properties have not changed)
+* It’s smaller than 64KB
```
flin-8, line 37: Worth putting in a link/explanation for smaller than 64KB?

OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,99 @@
...
+    "source": "Octofront",
+    "type": "SubscriptionCreatedV1",
+    "occurredTimeUtc": "2021-11-08T12:33:00Z",
+    "activityId": "00-15c3886877ecbd4a95b8b031cd7fbb46-c5fe780fb33bb443-00",
+    "specVersion": "1",
+    "payload": { }
+}
+```
+
+| Name        | Type        | Description | Implementation |
```
flin-8, line 89: This table isn't rendering properly atm

OctopusHQ/Events.md
pawelpabich: This file will be deleted soon

OctopusHQ/Strategy.md
pawelpabich: Sure, I just don't remember why we picked this number. @APErebus ?

OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,99 @@
...
+A good event has the following properties:
+
+* It’s treated as a strict contract that needs to be honoured for at least 6 months which translates currently to 2 3month longs cycles/quarters. In other words, it is unlikely to change.
+* It's frequency can be estimated reasonably well. E.g. 10s/100s/1000s a minute/an hour.
+* It’s specific (e.g. `SubscriptionTerminatedV1Event`, `OutageWindowChangedV1Event`) not generic (e.g. `SubscriptionUpdatedV1Event`)
+* Its entire payload is mandatory. E.g. All attributes have values assigned to them.
+* It only includes data specific to what it represents (e.g. `OutageWindowChangedV1Event` shouldn’t carry subscription limits just in case one of the consumers needs them)
+* It is published only when a real change has occurred (e.g. `OutageWindowChangedV1Event` should not be published if `Start` and `End` properties have not changed)
+* It’s smaller than 64KB
+* Its name ends with a version number and `Event` suffix. E.g. `OutageWindowChangedV1Event`
```
pawelpabich, line 38: I just had a look at the [API versioning doco](https://github.com/OctopusDeploy/Architecture/blob/0a5980e0918906425fc20087d8704b064e9a7053/OctopusServer/Api/ApiPhilosophy.md#api-versioning) for Octopus Deploy and their convention is to put the version number right at the end. I think that is a better approach. Should we adopt this convention? Did we have a specific reason to put `Event` at the end?

OctopusHQ/Strategy.md
APErebus: Hrm, can't remember off the top of my head. The standard Service Bus is 256kb message max with the headers taking up a max of 64kb...

OctopusHQ/Strategy.md
APErebus: Hrm, not sure. Might have been a side affect of how we were `OutageWindowChangedV1` but then we added the `Event` on the end...

My only concern is what do we do with our current events? Or just accept that for now they are "non-standard". Not sure we have the capacity to do any changeover/migration

OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,99 @@
...
+{
+    "id" : "SubscriptionCreated-licensekey-8a52c8af-2408-486f-ba4a-d277079de2b0",
+    "source": "Octofront",
+    "type": "SubscriptionCreatedV1",
+    "occurredTimeUtc": "2021-11-08T12:33:00Z",
+    "activityId": "00-15c3886877ecbd4a95b8b031cd7fbb46-c5fe780fb33bb443-00",
+    "specVersion": "1",
+    "payload": { }
+}
+```
```
APErebus, line 87: It looks like we've dropped the `path` field? Fwiw, I'm totally fine with that because I don't think it's being used properly or is serving any purpose

OctopusHQ/Strategy.md
pawelpabich: Thanks. I will add something

OctopusHQ/Strategy.md
pawelpabich: Thanks, it does render ok in VS Code. 🤔

OctopusHQ/Strategy.md
pawelpabich: Yeah, I had a comment that explained that but it looks like it was removed by subsequent changes.

OctopusHQ/Strategy.md
pawelpabich: @APErebus here it is :)

OctopusHQ/Strategy.md
```diff
@@ -0,0 +1,99 @@
...
+
+| Name        | Type        | Description | Implementation |
+| ----------- | ----------- | ----------- | ----------- |
+| id          | string      | The unique identifier | This property can be used to take advantage of [the message de-duplication feature offered by Azure Service Bus](https://docs.microsoft.com/en-us/azure/service-bus-messaging/duplicate-detection). At the minimum the identifier needs to be unique within the topic it belongs to.
+ |
+| source   | string        | The name of the source system | e.g. `Octofront`, `Cloud Platform`|
+| type   | string        | The type of the event | e.g. `OutageWindowChangedV1Event`|
+| occurredTimeUtc   | string        | An ISO 8601 formatted string representing the data and time of when the event occured  | |
+| activityId   | string        | An identifier that identifies the overall activity in traces (ToDo: We need more info here) | |
```
pawelpabich, line 96: @APErebus would be ok to explain here what this value should be? I don't think this description is usable as it is right now.

OctopusHQ/Strategy.md
pawelpabich: Ok, so we are ok to change the convention 👍.

Regarding the events we already have, let's use the process described here and let's rename them in the next 6 months.

OctopusHQ/Strategy.md
APErebus: Ah, ty :)

Yes, I'm all for dropping `path`. It didn't end up being necessary and the amount of work to support it properly is kinda prohibitive imo

OctopusHQ/Strategy.md
APErebus: Yeah, let me add something here

OctopusHQ/Strategy.md
APErebus: @pawelpabich Yeah, I'm ok with that.

APErebus (approved):
Looks good to me! Good stuff :)

flin-8 (approved):
LGTM

PR #68, opened 2022-02-17
Explained why hubs are  on LTS version of Octopus Server

pawelpabich:


CloudPlatform/DependencyManagement.md
```diff
@@ -20,6 +20,10 @@ Backend dependencies will be updated using the same approach as the one used for
 
 Cloud Platform is a glue that connects several external services (Azure SQL, Azure Storage Account, Azure Kubernetes Service, Better Uptime, Seq, SumoLogic, etc.). Some of them provide official libraries that wrap their APIs.  These libraries need also be kept up to date.
 
+# Hubs #
+
+[Hubs](https://github.com/OctopusDeploy/hosted-docs/blob/master/what-makes-up-cloud-platform/what-makes-up-cloud-platform.md) are dedicated Cloud instances that Cloud Portal uses to manage Cloud Platform. Cloud Portal can't effectively manage Cloud Platform without them being fully functional. This is why hubs are running the latest LTS (aka what's available on [the downloads page](https://octopus.com/downloads)) version of Octopus Server instead of running the latest code from master.
```
flin-8, line 25:
```suggestion
[Hubs](https://github.com/OctopusDeploy/hosted-docs/blob/master/what-makes-up-cloud-platform/what-makes-up-cloud-platform.md) are dedicated Cloud instances that Cloud Portal uses to manage the Cloud Platform. Cloud Portal can't effectively manage the Cloud Platform without the hubs being fully functional (imagine all Cloud instances being down). This is why hubs are running the latest LTS (aka what's available on [the downloads page](https://octopus.com/downloads)) version of Octopus Server instead of running the latest code from master.
```

flin-8 (approved):
LGTM

CloudPlatform/DependencyManagement.md
pawelpabich: English is not my first language but my understanding is that custom names like Cloud Portal, Cloud Platform don't require `the` as everybody knows what they are.  I checked Octofornt in a few places and it is never prefixed with `the`  🤔

flin-8: Not my first language either 😂 

You're right... `Cloud Platform` doesn't need "the".... I got confused because it's such a common term

kengel100: 🤷‍♀️

PR #69, opened 2022-02-22
Introduced Query API

pawelpabich:
One of the missing pieces in  OcotpusHQ is the ability to retrieve additional data when:
* an event doesn't have all the data that the consumer needs
* a new system joins OctopusHQ and needs access to historical data

This PR proposes a solution to this PR. All feedback is more than welcome!


pawelpabich:
@nathanwoctopusdeploy @APErebus it would be great if you could have a least quick look at this PR before our tomorrow catchup. We can review it there together if this helps.

OctopusHQ/Strategy.md
```diff
@@ -20,23 +24,36 @@ A the time of writing there are the following systems in Octopus HQ:
...
+* It’s treated as a strict, immutable contract that needs to be honoured for at least 6 months. 
 * It's frequency can be estimated reasonably well. E.g. 10s/100s/1000s a minute/an hour.
 * It’s specific (e.g. `SubscriptionTerminatedEventV1`, `OutageWindowChangedEventV1`) not generic (e.g. `SubscriptionUpdatedEventV1`)
-* Its entire payload is mandatory. E.g. All attributes have values assigned to them.
-* It only includes data specific to what it represents (e.g. `OutageWindowChangedEventV1` shouldn’t carry subscription limits just in case one of the consumers needs them)
-* It is published only when a real change has occurred (e.g. `OutageWindowChangedEventV1` should not be published if `Start` and `End` properties have not changed)
-* It’s smaller than 64KB. ASB can handle bigger messages. That being said, the maximum size of each topic is 3GB so this limit should be a good rule of thumb that will prevent us from using all available space.
+* It only includes enough information to identify the source of the event. E.g. `SubscriptionTerminatedEventV1` includes only `Id` of the subscription that has been terminated.
```
nathanwoctopusdeploy, line 38: I think events should still contain as much data as is relevant to the event and not be restricted to being slim, identification only properties. If a consumer can make all relevant decisions just on the event then it means no additional query api usage needed. If more data is needed, or additional data from other bounded context then the query api comes into play

OctopusHQ/Strategy.md
```diff
@@ -49,7 +66,8 @@ Every topic will have at least one subscription called `audit` which will store
 
 ## Security ##
 
-There will be a dedicated topic per bounded context. This should give us enough flexibility to implement effective security controls (e.g. who can send messages to a topic,  who can subscribe to a topic)  in the future. This work hasn't been done yet.
+There will be a dedicated topic per bounded context. This should give us enough flexibility to implement effective security controls (e.g. who can send messages to a topic,  who can subscribe to a topic)  in the future. This work hasn't been done yet but this the current plan.
+
```
nathanwoctopusdeploy, line 70: Query API security would be a good one to look at as well

OctopusHQ/Strategy.md
```diff
@@ -61,12 +79,10 @@ There will be a dedicated topic per bounded context. This should give us enough
 * Can I subscribe to an event as soon as it is announced? 
     * Yes, as long as the topic where the event will be published already exists. 
 * How do I get historical events/data?
-    * This problem doesn't have a good solution yet.
-    * Ideas: 
-         * Use public APIs provided by the system that publishes the event
-         * Request a data dump from the system that publishes the event
+    * Use the query API
```
nathanwoctopusdeploy, line 82: Do we want to look at a set of 'baseline' services each provider must expose to be able to get many things, a single thing etc.

OctopusHQ/Strategy.md
```diff
@@ -61,12 +79,10 @@ There will be a dedicated topic per bounded context. This should give us enough
...
+    * Request a data dump from the system that publishes the event
 * How do I notify the wider ecosystem about changes to the events I publish?
-    * Now: Update System Events section of this document and  announce your changes in `#topic-octopushq-integration`.
+    * Add your events and queries as JSON schema document to this repo and announce your changes in `#topic-octopushq-integration`.
```
nathanwoctopusdeploy, line 85: The ability for consumers to generate clients or event for a bounded context exposing a client library (think ms client libraries) would be something to look at to facilitate easy consumption of events / query apis

While it's falling into implementation details some guidelines may be good e.g. openapi etc.

OctopusHQ/Strategy.md
```diff
@@ -20,23 +24,36 @@ A the time of writing there are the following systems in Octopus HQ:
 
 ## Interface ##
 
-The current approach is a mix of [Event Notifications](https://martinfowler.com/articles/201701-event-driven.html#EventNotification) that is used for events representing state changes and [Event-Carried State Transfers](https://martinfowler.com/articles/201701-event-driven.html#Event-carriedStateTransfer) that is used for well-scoped data updates. As long as we avoid `TheWholeThingUpdated` events, this approach, which is not perfect, should be a good starting point.
+The current approach is based on [Event Notifications](https://martinfowler.com/articles/201701-event-driven.html#EventNotification) used to propagate information about significant state changes and query API for requesting the details of the change.
```
nathanwoctopusdeploy, line 27: It's probably more of an edge case but what about a consumers who need the exact state of a thing at the time an event is published about that thing. If we have minimal data in events then a consumer has to receive the event and perform a query which takes period of time. The thing could have changed after the event was published and before the query api was used to retrieve the things data.

OctopusHQ/Strategy.md
```diff
@@ -20,23 +24,36 @@ A the time of writing there are the following systems in Octopus HQ:
...
+
+### Implementation ###
+
+When a system receives an event it should be able to make a good enough decision based on the event itself and its internal data. This removes the query API, which often will require synchronous communication, from the critical path. E.g. When Cloud Portal receives `SoftwareProblemDetectedEvent` it will block all upgrades until it can retrieve additional information about the problem using the query API provided by the owner of `Release Management` context.
+
+
+More than likely, the query API will be a synch API which means that most of the event handlers will need to become long-running processes (aka [Saga](https://docs.particular.net/nservicebus/sagas/), [Background Task](https://github.com/OctopusDeploy/Architecture/blob/main/CloudPlatform/BackgrundProcessesAndTasks.md), [Business Process](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Core/Model/IBusinessProcess.cs), [Durable Azure Function](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-overview?tabs=csharp))
```
nathanwoctopusdeploy, line 54: sync (synch)

OctopusHQ/Strategy.md
```diff
@@ -61,12 +79,10 @@ There will be a dedicated topic per bounded context. This should give us enough
...
-    * This problem doesn't have a good solution yet.
-    * Ideas: 
-         * Use public APIs provided by the system that publishes the event
-         * Request a data dump from the system that publishes the event
+    * Use the query API
+    * Request a data dump from the system that publishes the event
 * How do I notify the wider ecosystem about changes to the events I publish?
-    * Now: Update System Events section of this document and  announce your changes in `#topic-octopushq-integration`.
+    * Add your events and queries as JSON schema document to this repo and announce your changes in `#topic-octopushq-integration`.
     * Future: Some form of contract repository
```
nathanwoctopusdeploy, line 86: That's an interesting one. When it comes to query api I lean towards a 'live' contract rather than a documented one which may be out of date or the contact repository needs to be auto generated so it matches the actual implementation. I've seen a lot of documented apis / events that don't actually match the implantation

OctopusHQ/Strategy.md
APErebus: Yeah, I agree with this.

OctopusHQ/Strategy.md
APErebus: I think in this particular case it makes sense to have a more strict contract. That being said, I think having the contract be "documented" via Swagger/OpenAPI (in the case of HTTP REST), is probably easier on the devs and maybe only a changelog needs to be documented.

APErebus (commented):
Broadly I think this is a good approach! 👏

OctopusHQ/Strategy.md
pawelpabich: How would we know what data to include? This is relatively easy to answer for `OutageChangedEvent`  but what would you include in SubscriptionCreateEvent without interviewing all possible consumers?

OctopusHQ/Strategy.md
pawelpabich: Agreed. I was thinking about API keys via Bearer Token.

OctopusHQ/Strategy.md
pawelpabich: That might introduce accidental complexity. E.g. What we need is GetById API but we will also expose GetAll API which then required paging.

OctopusHQ/Strategy.md
pawelpabich: Good question. Do you have a specific example in mind?

OctopusHQ/Strategy.md
pawelpabich: We need a single source of truth for the contracts. I see Swagger as an easy way to generate clients but I'm not sure how we would detect drift.  "Devs will just do the right thing" doesn't feel good enough here :)

OctopusHQ/Strategy.md
pawelpabich: Reverting this change in favour of self-contained events as pros outweigh cons.

OctopusHQ/Strategy.md
pawelpabich: Reverting this change in favour of self-contained events as pros outweigh cons.

OctopusHQ/Strategy.md
pawelpabich: Will worry about that in the future

PR #73, opened 2022-03-30
Extract common logging guide and add log levels

flin-8:


skermajo (approved):
Awesome! Thanks for restructuring and adding the log levels. LGTM 👍

PR #75, opened 2022-04-12
Standardise on PascalCase naming convention

pawelpabich:
**Problem**

The lack of a standardised naming convention ([1](https://octopusdeploy.slack.com/archives/C02URML7HND/p1647820992791509), [2](https://octopusdeploy.slack.com/archives/C02URML7HND/p1649729417705449)) caused a bit of confusion so let's define it and solve the problem so the following "just in case" approach is not needed.

image
 
At the moment we use `PascalCase` naming convention for property names and `flatcase` naming convention for event names ([Octofront](https://github.com/OctopusDeploy/Octofront/blob/4d7af06b3d8543780bd034432c698b1645fd85d8/source/CloudPlatform.Events/OctopusEventPublisher.cs#L23), [Cloud Portal](https://github.com/OctopusDeploy/CloudPortal/blob/8ccf5530f728c3bde72fc0e40408be6915917d4f/source/CloudPortal.Events/OctopusEventPublisher.cs#L23)).

**Solution**

I propose we stick to `PascalCase` naming convention for everything (events, query API, property names). In this way, we have only one convention to follow.

This is a breaking change for `Purchasing`, `CloudPlatform`, and `ReleaseManagement` topics and the owners of all of those topics should migrate to then new structure in X months. 

No matter what we agree on, we also have to agree on how and when these changes will be rolled out so we don't have multiple different ways of doing the same thing which will only confuse more people in the future.

**Migration process**

0. Agree on START_DATE
1. Subscribers: Repalce `type='xeventv1'` with ``type='XEventV1'` or `type='xeventv1'`` by START_DATE + 2 weeks
2. Publishers: Repalce `type='xeventv1'` with `type='XEventV1'` by START_DATE + 4 weeks
3. Subscribers (optional but recommended): Repalce `type='XEventV1'` or `type='xeventv1'` with `type='XEventV1'`

Thoughts?


mcasperson:
👍 PascalCase makes sense to me for everything.

What was the idea behind changing `type='eeventv1'` to `type='XEventV1'`? Does the leading `X` have significance?

matt-richardson:
Should we use this as an early / simple test of our message versioning? ie, change to `V2`?
(not against the plan as proposed - just food for thought)

nathanwoctopusdeploy:
PascalCase sounds good 👍

pawelpabich:
> What was the idea behind changing type='eeventv1' to type='XEventV1'? Does the leading X have significance?
This is just my typo :). `eeventv1` => `xeventv1`

pawelpabich:
> Should we use this as an early / simple test of our message versioning? ie, change to V2?

Good question.  This is more a change to the protocol so I'm not sure a new version of events would help us here 🤔.  Do you have a specific approach in mind that you could describe?

matt-richardson:
> Do you have a specific approach in mind that you could describe?

Would it not be a case of:

1. Publishers: flag `xeventv1` as deprecated and start publishing `XEventV2` (as well as v1 for x months)
2. Subscribers: update code that looks for 'xeventv1' to look for 'XEventV2'`
3. Publishers: Drop `xeventv1` messages in x months

Seems a good use of versioning to me? I suspect that I'm missing something though.

pawelpabich:
There are two major parts in OcotpusHQ. Contracts (e.g. events) and protocol (how events get communicated).  The event names have been using `PascalCase` convention since day 1.  v1 of the protocol had a rule that wasn't reflected in the doco properly. The rule was that `type` property will use `flatcase` convention. 

This is what the doco should have had.
image

So technically speaking events haven't changed, it's the protocol that delivers them that has changed.  We should bump the protocol version to `2` to make that clear.

Happy to zoom if that easier.

PR #77, opened 2022-04-27
Add telemetry proxy architecture document

shaunmarx:
We recently made a whole bunch of changes to our telemetry proxy and octopus server to improve our telemetry proxy architecture, specifically to make it more resilient and allow us to improve how we generally send events to amplitude. This PR retrospectively adds the decisions captured.


TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+We should make it as simple as possible to send events for debugging purposes and run the telemetry proxy locally.
+
+**Scalable**
+
+We would like to enable the telemetry proxy for our self-hosted customers as well, but in order to do so we need a solution that can handle a large volume of events and scale appropriately. 
+
+
+## Solution
+
+We will use an `EventsHub` for our telemetry, which acts as both a queue and a mechanism to recover from failures as messages can be retained for 7 days, but can also
```
TomPeters, line 24:
```suggestion
We use an `EventsHub` for our telemetry, which acts as both a queue and a mechanism to recover from failures as messages can be retained for 7 days, but can also
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+Our telemetry proxy should be resilient to failures and be able to recover events if necessary. This also includes amplitude throttling events. As such, the telemetry proxy should transparently handle batching to amplitude as part of this process as well.
+
+**Developer Experience**
+
+We should make it as simple as possible to send events for debugging purposes and run the telemetry proxy locally.
+
+**Scalable**
+
+We would like to enable the telemetry proxy for our self-hosted customers as well, but in order to do so we need a solution that can handle a large volume of events and scale appropriately.
```
TomPeters, line 19:
```suggestion
Our telemetry proxy should be able to ingest traffic from both cloud hosted and self hosted customers, but in order to do so we need a solution that can handle a large volume of events and scale appropriately. 
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Constraints
+
+**Resilient**
+
+Our telemetry proxy should be resilient to failures and be able to recover events if necessary. This also includes amplitude throttling events. As such, the telemetry proxy should transparently handle batching to amplitude as part of this process as well.
+
+**Developer Experience**
+
+We should make it as simple as possible to send events for debugging purposes and run the telemetry proxy locally.
```
TomPeters, line 15:
```suggestion
It should be as simple as possible to send events for debugging purposes and run the telemetry proxy locally.
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+## Solution
+
+We will use an `EventsHub` for our telemetry, which acts as both a queue and a mechanism to recover from failures as messages can be retained for 7 days, but can also
+enable event capturing to store events indefinitely. We will then use an [azure function events hub trigger](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-hubs?tabs=in-process%2Cextensionv5&pivots=programming-language-javascript) to process the data separately. This provides us with a number of
+good characteristics:
+
+* Our throughput and limits are higher than what Amplitude can effectively provided as we can capture the data asynchronously
+* We can replay events, pushing said events to different amplitude projects
```
TomPeters, line 29:
```suggestion
* We can replay events, pushing said events to different amplitude projects, or recover events in some failure cases
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+* It allows us to push events to multiple destinations other than amplitude e.g. a data lake
+* We can deploy updates to our processing of events separately from our ingress endpoint
+* We can scale our ingress and processing separately
+
+In addition to the above, we will also capture additional metadata as part of our events to ensure we can push more responsibilities and better "routing" capabilities into our telemetry proxy. We will include the following additional user properties:
```
TomPeters, line 34:
```suggestion
We capture additional metadata as part of our events to ensure we can push more responsibilities and better "routing" capabilities into our telemetry proxy. We include the following additional user properties:
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+* License Hash
+* Hosting Environment (i.e. self hosted vs cloud)
+* Product (i.e. Octopus)
```
TomPeters, line 38:
```suggestion
* Product (e.g. Octopus Server. This field allows other products to use the telemetry proxy, and for the telemetry proxy to route these events to the appropriate Amplitude project)
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+* Version (Octopus version)
+* IsDebug 
+
+### Response Behavior
+
+Since we will have multiple code paths active at any point we will combine the results in such a way that if one of the code paths is successful, the response is a success. This is the case as any event directly logged to amplitude will have been appropriately captured, while any event pushed to the events hub can at any point be replayed under the right conditions such as retention period of events not exceeded if storage capturing isn't enabled.
```
TomPeters, line 44: I don't think we need to talk about the multiple code paths, as this is an implementation detail (and a short term one).

```suggestion

From the ingest endpoint, the following responses can be returned
- `200` if the events were successfully ingested. The event is then guaranteed to end up in Amplitude.
- `429` if there were too many requests and the client needs to retry after a delay. This should happen very rarely, as the ingestion service should scale up in this scenario.
- `40x` or `50x` if something went wrong. In this case, it is unlikely that retrying the ingestion request would cause it to succeed, so it is recommended that these events be dropped.

It is acceptable to send the same event multiple times to the ingestion endpoint as it will be de-duplicated.

If an event was accepted by the ingestion endpoint, it is not guaranteed to end up immediately in Amplitude as it will first be throttled and batched by the EventsHub trigger.

```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Backwards Compatibility 
+
+We need to ensure that we don't break our existing implementation of pushing events directly to amplitude so we both our new proposed events hub integration as well as the old code path will exist side by side. These paths will be turned on and off using environment variables. We should also make sure that our routing takes into account that we will be receiving a number of events from older instances which will not have the new user properties such as `Product` and `Hosting Environment`. In these cases we should default to pushing to our fallback amplitude project, which in our case would be the production amplitude project as all existing events would be coming from cloud instances.
```
TomPeters, line 48: I think you can remove this section

```suggestion
```

TelemetryProxy/Architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Routing and filtering
+
+We will initially only forward 25% of self-hosted customer events, which can be calculated from the `License Hash` and `Hosting Environment` properties being introduced. We will also ensure that when the `IsDebug` user property is provided, we will route events to our `Dev/Test` amplitude project which will remove the need for developers to jump through hoops to send and test events.
```
TomPeters, line 52:
```suggestion
We forward a percentage (e.g. 25%) of self-hosted customer events to amplitude, which can be calculated from the `License Hash` and `Hosting Environment` properties. This is done to limit the cost of ingesting a large number of events in Amplitude, but should still give us enough insights into how customers are using the product. This can be easily scaled up or down.

If the `IsDebug` user property is provided, we route events to our `Dev/Test` amplitude project which removes the need for developers to jump through hoops to send and test events.
```

TomPeters (approved):
This looks pretty good.

I've left some inline comments that roughly correspond to these two points:

- Try to write this document more like a statement of how the system works, rather than an intention of what we had planned to do (that reads more like a proposal).

- I think some of the details you've put in here are more like implementation details, rather than part of an architecture document.

PR #78, opened 2022-04-28
Add sensible defaults for using build containers

cailyoung:
Initial draft of the guidance to place in here. All feedback welcome.


SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+Suits best if the build container is small (we don't want our teams to be waiting a long time for their build container to be built before their application has even started building) and to make specific atomic changes to the build environment and build process and application being built.
+
+This pattern does not cache the build environment anywhere outside the build agent running your overall build.
+
+We do not recommend using this pattern if you are anticipating that the container build will take a long time versus pulling the container (e.g. if you are compiling something from source in the container build). Use the next pattern if your container build is slow.
+
+### Build container built seperately to application build
+
+Suits best if the build process is slow, or your team doesn't want to take on the cognitive load of having a completely ephemeral build container as part of your build process.
```
TomPeters, line 58: In both of these documents
- https://octopushq.atlassian.net/wiki/spaces/RND/pages/2388623576/Define+the+build+environment+as+a+Dockerfile+build+it+in+your+build+config+as+the+first+step
- https://octopushq.atlassian.net/wiki/spaces/RND/pages/2388231045/Stand+up+a+new+build+pipeline+to+build+your+Dockerfile+push+it+into+the+private+registry

You mention using these approaches if your team can handle the cognitive load of these approaches. Implying that there is a different kind of cognitive load to take on for each of them. Consider adding more details here, or using different language rather than "cognitive load" to describe the tradeoff, to make it clearer to people about the tradeoff they might be making.

For example
- Containers built separately to the application build incur additional complexity in the build pipeline, as there needs to be more build configurations, the container needs to be built and updated regularly.
- Containers built inline with the CI application build need to tag images in a certain way to ensure uniqueness, which could potentially cause confusion during local dev

Or whatever the actual tradeoff is :)

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+# Antipatterns
+
+## (Nuke-specific) Wrap everything in a `DockerTasks.DockerRun`
+
+We tried this for TentacleArmy, that used a nuke build script to run its application build and noticed that you lose _all_ of the advantages of of nuke's fluent API wrapping of the `dotnet` (and others) commands. The environment also had to ensure that docker cli was runnable which meant you were unable to use official Microsoft-provided .NET SDK containers to run. If you are running into this build antipattern consider instead using `tool-docker-cli` and using a single build step. 
+
+## Use a Docker container as a way to cache packaged application dependencies
+
+If you're considering putting `dotnet restore` in a custom build container to speed up compilation, please don't. Here's why:
+- If your custom container doesn't get rebuilt/re-pushed/re-pulled, then any security fixes in third-party dependencies will not be included in your actual build. This can be very tricky to debug, depending on the package ecosystem and how they handle cache resolution
```
TomPeters, line 73: From my understanding, the effect of the dotnet restore is to build up .net's package cache. The actual build still does a `dotnet restore`, ensure that it gets updated dependencies (i.e. there shouldn't be any security issues), but because of the pre-populated package cache, most of the requests to download packages hit the cache instead of going to nuget.org, or feedz, or wherever.

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+- The time spent pushing/pulling/expanding/copying very large container layers isn't materially faster than the time pulling direct from external package sources, _even accounting for occasional unavailability_.
```
TomPeters, line 74: Do we have any data on this? I believe we currently use this approach based on some data that andrew h gathered in the past, so if we have data that contradicts that, then that would be interesting to explore further :)

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+  - If you are finding external packages sources are unreliable, please let #team-engineering-productivity know and we can see about enabling caching within our build cluster for your use case
```
TomPeters, line 75: I would emphasise this point.

Speeding up package restoration for all projects would eliminate toil across teams, and a bunch of extra work that teams might need to otherwise do to create pre-populated container images, and would also provide a better service (since all packages would get cached, rather than just the ones captured the last time the container image was built)

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+> We acknowledge that `build-base-dotnet` currently includes a large NuGet and npm package cache, and it is our analysis of our customer's use of that container that leads us to recommend against using this pattern in general. We explicitly do not recommend using this container for any new projects.
```
TomPeters, line 77: I'd leave out this bit - it feels a bit confrontational ("This customer is doing it bad - don't be like that customer").

```suggestion
> We acknowledge that `build-base-dotnet` currently includes a large NuGet and npm package cache. We aim to remove this container as a base container for our teamcity agents, and we explicitly do not recommend using this container for any new projects.
```

TomPeters (commented):
One thing that feels missing here is a discussion about what we will do with workloads that need to run on windows. If we don't have an answer yet, then we should at least say that for now.

After thinking about this for a bit, there are a few other patterns here that I think would be good to explore, even if it is just to explicitly rule them out.

## 1: Using containers as artifacts

You could have a docker file that `RUN`s the build script (e.g. `build.sh`). The result of building the project is then captured as a layer in the container.

This docker container could be pushed to artifactory.

A subsequent build step, or a different build configuration, could pull that docker container image and run tests within it.

This might work well for something like Octopus Server. Some benefits:
- The only build configuration / build step that needs to know about tools (like docker/nuke/.net/etc.) is the main build step/configuration. Subsequent steps just pull the container from the previous step.
- We don't need to push artifacts to/from teamcity - everything sits in our container registry.

## 2: Docker multi-stage builds

If you have a few different build steps that have different build requirements, the current recommendation feels like it is to run each of those build steps in a different container.

Can we use multi-stage docker builds to run it in a single docker build steps instead?

E.g. if you had a build that is like

1. Build frontend
2. Build server
3. Bundle together
4. Push to Octopus Deploy
5. Create release

You could do something like

```
FROM node:latest:latest AS build-frontend
...
RUN npm run build
COPY ...

FROM mcr.microsoft.com/dotnet/sdk:latest AS build-server
...
RUN ./build-backend.sh
COPY ...

FROM mcr.microsoft.com/dotnet/sdk:latest AS bundle
COPY --from=build-frontend ./ ./frontend
COPY --from=build-server ./ ./server
RUN ./bundle.sh

FROM octopusdeploy/octo:latest as push
COPY --from=bundle ./ ./
RUN create-release ...
```

You could run any of these stages at different parts of the build configuration.

Something along these lines might make it easier for any builds that need multiple tools. For example, the [tentacle army build](https://github.com/OctopusDeploy/TentacleArmy/blob/master/ci-environment.Dockerfile) is effectively composing the docker-cli container and the off the shelf .net sdk container, but it can't just re-use the .net sdk container easily. Is there a way we can use the public .net sdk container?

In some ways it feels like we can, but only if we take away some of the responsibility of nuke to manage dependencies between build tasks.

Really curious for some to give this some :brain: time.

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+- Understand and own your own dependencies
+
+- Isolate your environment from other teams
+  - Dependency changes
+  - Inter-build pollution
+
+# Patterns
+
+## No Container :raised_eyebrow:
```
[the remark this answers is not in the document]
casstait: We can move this down the list to de-emphasize :) sure thing.

SensibleDefaults/BuildInContainers.md
casstait: Thanks for the feedback! The "cognitive load" we were implying here was a bit on the extra build configurations but also on the cognitive load of developing an image/Dockerfile that would need to be maintained and developed appropriately (i.e. not be 30GB large). We can clear that up to delineate these two scenarios.

casstait:
@TomPeters Thanks for the extra comment/feedback! Its super helpful for us to progress.

Can I clarify a few things with you though, as I'm not entirely clear on some reasons why. 

> One thing that feels missing here is a discussion about what we will do with workloads that need to run on windows.

We were trying to be build platform and I guess in affect OS agnostic in our documentation. What special details or extra information were you wondering about with workloads that need to run on windows opposed to say linux? 

> 1: Using containers as artifacts

I'm not entirely sure on this approach for a couple of reasons:
1. This would create an infinite number of images that would be pushed to artifactory and need to have a retention policy against to ensure infinite growth/storage doesn't occur. 
2. I don't believe you need a seperate image for your build that could then be referenced for your tests. You can absolutely use the same container to build your application and then subsequently run tests against that same container, eliminating the issues discussed in 1.
3. The way to perform 2. from the top of my head would have to be performed on the same agent and therefore in the same build configuration. 
4. Would could potentially push a unique image to artifactory if there is a final build step of removing that image once the build chain is complete. (options!)

> 2: Docker multi-stage builds
>
> If you have a few different build steps that have different build requirements, the current recommendation feels like it is to run each of those build steps in a different container.

That is the current recommendation yep, if we want our build steps to be small simple tasks that can be run in the smallest and least complex docker container

> Can we use multi-stage docker builds to run it in a single docker build steps instead?

I believe we can and something that Cail and I should explore further on implementing appropriately in TeamCity.

> For example, the [tentacle army build](https://github.com/OctopusDeploy/TentacleArmy/blob/master/ci-environment.Dockerfile) is effectively composing the docker-cli container and the off the shelf .net sdk container, but it can't just re-use the .net sdk container easily. Is there a way we can use the public .net sdk container? 

Tentacle Army as the example (and any other nuke builds that have `DockerTasks`) can't use the public .net sdk container because it doesn't have docker installed :(. So that's why we have suggested the `tool-docker-cli` tool and anything that needs docker. For nuke especially, there is a boostrapper script that installs the required .net sdk to run nuke, so that's where this line of thought went.

SensibleDefaults/BuildInContainers.md
cailyoung: Many builds use `--no-restore`, according to my 'anecdata' here. It's a pattern I've seen in at least one build, anyway. However, thinking about the example given here, I think it's a bit contrived, and am inclined to remove it.

SensibleDefaults/BuildInContainers.md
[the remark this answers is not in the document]
cailyoung: I actually put it at the top on purpose, as it is the 'status quo' approach, and I think for an audience that isn't already using a container, it might be helpful to show the limited circumstances we think that's appropriate before moving on to the others.

However, if that's not the preferred ordering here, I'm not going to insist it stays where it is.

SensibleDefaults/BuildInContainers.md
TomPeters: > Many (compilation) builds use --no-restore

Really!? Can you share any with me so I can have a look?

cailyoung: https://github.com/OctopusDeploy/OctopusDeploy/blob/d5c3fa4f2ff21629ba3dc605eac9c64017de99ee/nuke-build/Publish/AssemblyPublisher.cs#L45

cailyoung: Although the git blame on that one is... 😬 😉

TomPeters:
> What special details or extra information were you wondering about with workloads that need to run on windows opposed to say linux?

For example, we recommend that 

> If your build needs to talk to a Docker daemon itself, but has no other dependencies (or can handle bootstrapping itself), use this container: `docker.packages.octopushq.com/octopusdeploy/tool-containers/tool-docker-cli:latest`

If your build also needs to run windows, this isn't quite right because that's a linux container image.

SensibleDefaults/BuildInContainers.md
TomPeters: Without following the chain too deeply... The build configuration in teamcity still does a restore right? By virtue of whatever target calls that function depending on a different target that does a restore. So as long as the restore still happens as part of the build configuration, then happy days. Just asking to make sure I'm not missing anything :)

TomPeters:
> I'm not entirely sure on this approach for a couple of reasons

Great! I think it is worth exploring the topic, thinking about the pros and cons, and writing down these reasons in this document :)

SensibleDefaults/BuildInContainers.md
cailyoung: 🤦 You're right. I'm misremembering a chain of events. The possibility is still there, but perhaps I'm conflating a bunch of disparate things I've been associated with.

TomPeters:
> Tentacle Army as the example (and any other nuke builds that have DockerTasks) can't use the public .net sdk container because it doesn't have docker installed :(. So that's why we have suggested the tool-docker-cli tool and anything that needs docker. For nuke especially, there is a boostrapper script that installs the required .net sdk to run nuke, so that's where this line of thought went.

As I see it, the problem is that we have nuke orchestrated docker. Which means we need to have docker + .net sdk available in the container.

I think this suggestion is about inverting that relationship. So the orchestration of (dependencies between) our build steps happens in a docker file instead of being in nuke, then nuke doesn't need to run docker.

So I'm challenging the assumption that nuke builds should be running DockerTasks.

SensibleDefaults/BuildInContainers.md
casstait: @TomPeters have updated those two confluence pages. Hopefully they read a bit better on the difference between the complexity between those two scenarios.

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+This is only a reasonable choice if the build commands are _so_ simple they are guaranteed to be available on _any_ base agent environment. The contents of the base agent environment is subject to change over time, and even tools like `curl` aren't guaranteed to stay available. Strongly consider using a container, even if it's something like `busybox` (which offers a lot of tools in a very small payload)
+
+## \[Recommended\] Public Container :earth_asia:
+
+If possible, use a publically-available container that already has your dependencies. This avoids the need for you to build and maintain your own.
```
matt-richardson, line 23: Strongly recommend some guidance to use a container from reputable source here.
(I recall I've probably already mentioned this, but I still feel strongly about it :smile:)

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Examples
+- `mcr.microsoft.com/dotnet/sdk:<version>` for .NET SDKs
+
+## Custom Agent Environment :hammer:
+
+If your build _cannot_ be run in a container you'll need to define or select a specific agent type that already has what you need. _More on this to be written..._
```
matt-richardson, line 30: I think more needs to be written here :)

matt-richardson: > select a specific agent type that already has what you need

As we're aiming for to deprecate as many of these as possible, it's probably wise to mention that and suggest that if at all possible, we should aim to rearchitect so we can run in a container (albeit with the caveat that we need to test things that are old). (I'm thinking of the "we need windows to run the code signing" / "nah, just use a different tool" conversation that was had)

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Examples
+- The Windows Server / SQL Server agents available in TeamCity
+
+## `tool-docker-cli` as the environment
+
+If your build needs to talk to a Docker daemon itself, but has no other dependencies (or can handle bootstrapping itself), use this container.
```
matt-richardson, line 37: Given it's likely we'll need to stop running privileged containers, should we mention that here?
(I think I might be conflating the "run child containers" and "run sibling container" situations)

SensibleDefaults/BuildInContainers.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Defined build environment in a Dockerfile
+
+Specify your build environment in its own Dockerfile in your repo, and build it as part of your build pipelines. There are two approaches here - either you build it on every build you run, or you build it infrequently and treat it as another versioned dependency.
+
+### Build container built inline with CI application build 
+
+Suits best if the build container is small (we don't want our teams to be waiting a long time for their build container to be built before their application has even started building) and to make specific atomic changes to the build environment and build process and application being built.
+
+This pattern does not cache the build environment anywhere outside the build agent running your overall build.
```
matt-richardson, line 52:
```suggestion
This pattern does not cache the build environment anywhere outside the build agent running the build.
```

SensibleDefaults/BuildInContainers.md
matt-richardson: I think it's worth adding the "cognitive load" reasoning to this document (as it's trying to explain the why)  as well.

matt-richardson (commented):
Finally had a chance to review. I've added a few cheeky comments, but also some more serious ones.
There's "main thread" comments that I will add subsequently to this.

matt-richardson:
> ## 1: Using containers as artifacts
> You could have a docker file that RUNs the build script (e.g. build.sh). The result of building the project is then captured as a layer in the container.
> 
> This docker container could be pushed to artifactory.
> 
> A subsequent build step, or a different build configuration, could pull that docker container image and run tests within it.
> 
> This might work well for something like Octopus Server. Some benefits:
> 
> * The only build configuration / build step that needs to know about tools (like docker/nuke/.net/etc.) is the main build step/configuration. Subsequent steps just pull the container from the previous step.
> * We don't need We don't need to push artifacts to/from teamcity - everything sits in our container registry.

You often want multi stage builds docker here. Subsequent steps should have the cruft (such as the .net sdk) that is needed by earlier steps.

The other aspect here is that you're effectively creating a big fat zip file, but it's an "opaque" docker file rather than a zip file. So in some ways, you're getting benefits from the parallel layer upload/download, and avoiding slow zip/unzip times, but in other ways you're just trading them against a "less portable" archive. (Not a pro, not a con, just something to consider).

Also, pushing round these large images isn't really much fun. We've tried to improve it, but while it's not hideous at the moment, our docker image upload/download isn't fast. (We've got multiple cards on our board about docker caching...)

matt-richardson:
> > Tentacle Army as the example (and any other nuke builds that have DockerTasks) can't use the public .net sdk container because it doesn't have docker installed :(. So that's why we have suggested the tool-docker-cli tool and anything that needs docker. For nuke especially, there is a boostrapper script that installs the required .net sdk to run nuke, so that's where this line of thought went.
> 
> As I see it, the problem is that we have nuke orchestrated docker. Which means we need to have docker + .net sdk available in the container.
> 
> I think this suggestion is about inverting that relationship. So the orchestration of (dependencies between) our build steps happens in a docker file instead of being in nuke, then nuke doesn't need to run docker.
> 
> So I'm challenging the assumption that nuke builds should be running DockerTasks.

If we go back to an "idiomatic use of docker", then we regress in some ways, and introduce other issues:

* we regress in that we lose the benefits of nuke (language that most of our devs are pretty damn good at, type safety, refactoring etc) and go back to the dark days (similar to the pain we had with cake)
* we have hard evidence (workertools) that docker files are easy to do badly, if the teams doing them are not experienced.

I'm not keen to lose the benefits of nuke orchestration here (where it makes sense (ie, for dotnet projects)).

TomPeters:
> If we go back to an "idiomatic use of docker", then we regress in some ways ....

I think this is a bigger, ongoing discussion that we should have and we should take outside of this PR.

TomPeters:
> The challenge is trying to get a bunch of these to run in parallel for speed. I like to think that the way we've setup the server builds is probably close to the fastest "time to start" that we can get.

I was thinking about this and a related problem: What if you want to push multiple docker images from a single build step?

I think the solution to this is probably the same solution to the parallelisation problem: Use docker compose instead of multi-stage docker builds.

PR #82, opened 2022-05-11
Add guidance about not requiring more than one environment

uglybugger:
https://octopusdeploy.slack.com/archives/C01MSG3LZ5M/p1652232126761519


mjhilton (approved):
Thanks for the clarification, LGTM

